PR #6, opened 2025-05-07
[Feature] Add main graph object

jpmoutinho:


docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
...
+
+R = uniform(
+    low = graph.max_distance(connected = True),
+    high = graph.min_distance(connected = False)
+)
+
+print(graph.ud_edges(radius = R) == graph.sorted_edges)
+```
+
+We can also reset the set of edges on a graph to be equal to the set of unit-disk edges for a given radius with `graph.set_ud_edges(radius = R)`. For this example we will not run this line, but it could be useful when constructing graphs from sets of coordinates.
```
nataliejpg, line 167: I like this concept, I would suggest that it be possible (and documented) to construct a graph from a set of nodes and coordinates and then set the edges like this as an alternative construction.

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
...
+If the graph does not have these attributes, the dictionaries will still be returned with `None` in place of the value.
+A set of boolean properties allows quickly checking if the graph has coordinates or weights. It only returns `True` if there is a value set for every node / edge in the graph.
+
+!!! note
+    Accessing weights as NumPy arrays or Torch tensors will be designed later.
+
+```python exec="on" source="material-block" session="graphs"
+assert graph.has_coords
+assert graph.has_node_weights
+assert graph.has_edge_weights
```
nataliejpg, line 190: I'm wondering if it would be good to also have easy access to the `ud_radius_range` s.t if a graph `is_ud_graph` you get back the low and high possible radius values rather than having to know to check for `is_ud_graph` and then call `graph.max_distance(connected = True)` and `graph.min_distance(connected = False)` but that may be overengineering things

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
+# Standard structure for Graphs
+
+Working with graphs is an essential part of computations with the Rydberg-Ising analog model. For that reason, QoolQit implements a specific `BaseGraph` class to serve as the basis of all graph creation and manipulation, and setting the logic related to unit-disk graphs. QoolQit integrates with [NetworkX](https://networkx.org/) for many operations, and the `BaseGraph` inherits from `nx.Graph`.
+
+The `BaseGraph` is meant as a base class, although currenty not abstract, and is not meant to be used directly. Instead, for data manipulation the user should rely on the `DataGraph`, which inherits from the `BaseGraph`, and defines specific constructors and extra logic to deal with node and edge weights.
+
+
+## Basic construction
+
+The `DataGraph` is an undirected graph with no self loops. Like a `nx.Graph` it can be instantiated empty and nodes or edges added afterwards. However, currently this is not advised. Instead, the default way to instantiate a `DataGraph` should be directly with a set of edges.
```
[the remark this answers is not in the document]
nataliejpg: Having read further down that you also support `from_coordinates` I'd suggest to mention this here (by saying, "one way to instantiate a graph is by its edges, see xxx section for instantiation by coordinates"

docs/contents/graphs.md
nataliejpg: So having read further down that it is I'm leaving this comment just to illustrate that if you read the docs from the top it seems like it isn't possible to do this when really you have already implemented it so it's not the code that needs to change, just the docs.

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
...
+# An Erdős–Rényi random graph
+graph = DataGraph.random_er(n = 10, p = 0.5)
+
+graph.draw()
+fig = graph.draw(return_fig = True) # markdown-exec: hide
+plt.tight_layout() # markdown-exec: hide
+print(fig_to_html(fig)) # markdown-exec: hide
+```
+
+### Loading from a matrix
```
nataliejpg, line 250: This section is quite unclear. Does this imply that a graph loaded in this way will be fully connected? Is there a use case for this? It seems close to QUBO but not exactly. My instinct would be to leave it out of this version until the need and use case is more clearly defined

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
+# Standard structure for Graphs
+
+Working with graphs is an essential part of computations with the Rydberg-Ising analog model. For that reason, QoolQit implements a specific `BaseGraph` class to serve as the basis of all graph creation and manipulation, and setting the logic related to unit-disk graphs. QoolQit integrates with [NetworkX](https://networkx.org/) for many operations, and the `BaseGraph` inherits from `nx.Graph`.
+
+The `BaseGraph` is meant as a base class, although currenty not abstract, and is not meant to be used directly. Instead, for data manipulation the user should rely on the `DataGraph`, which inherits from the `BaseGraph`, and defines specific constructors and extra logic to deal with node and edge weights.
```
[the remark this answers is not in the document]
jpmoutinho: Fair enough, this becomes more relevant once I add the `LayoutGraph`, but I'll see about the best way to mention it then.

docs/contents/graphs.md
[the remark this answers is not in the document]
jpmoutinho: Yes I had just done a quick mention for now, but essentially it related to this issue that I just created: https://github.com/pasqal-io/qoolqit/issues/24

For now I just removed the sentence, and added the link to the constructors as well.

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
...
+```
+Both a dictionary or a list can be passed, which will be converted to a dictionary. Because the graph now has a set of node coordinates, we can directly access the distance between the nodes. Optionally, a set of node pairs can be given and only those distances will be computed.
+
+```python exec="on" source="material-block" result="json" session="graphs"
+# Compute for all node pairs
+graph.distances()
+print(graph.distances()) # markdown-exec: hide
+
+# Compute only for connected nodes
+graph.distances(graph.sorted_edges)
```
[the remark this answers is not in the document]
jpmoutinho: I think it's clear, the `sorted_edges` property is documented earlier.

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+Graph coordinates can be rescaled.
+
+```python exec="on" source="material-block" session="graphs"
+# Rescale coordinates by a constant factor
+graph.rescale_coords(scaling = 2.0)
+
+# Rescale coordinates by setting the minimum spacing
+graph.respace_coords(spacing = 1.0)
```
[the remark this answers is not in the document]
jpmoutinho: Also a valid design but requires more condition handling. Personally I find the separation in different functions cleaner. Well both methods are documented next to each other so I don't think that is a problem.

docs/contents/graphs.md
```diff
@@ -0,0 +1,284 @@
...
+
+!!! info "Definition: Unit-Disk Graphs"
+    For a set of nodes $V$, each node $i\in V$ marked by a set of coordinates $(x, y)_i$ in Euclidean space, a set of edges $E$ and a radius $R$, a Unit-Disk Graph $UDG(V, E, R)$ is such that there exists an edge $(i, j)\in E$ for two nodes $i$ and $j$ if and only if $\text{dist}(i, j) \leq R$, where $\text{dist}(i, j)$ is the Euclidean distance between the coordinates of nodes $i$ and $j$.
+
+In other words, a unit-disk graph for a radius $R$ is the graph where the set of edges corresponds to the intersections of disks of radius $R$ centered at each node position in Euclidean space.
+
+For a `DataGraph` with a set of node coordinates, we can check if it is a valid unit-disk graph
+
+```python exec="on" source="material-block" result="json" session="graphs"
+graph.is_ud_graph()
```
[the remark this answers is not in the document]
jpmoutinho: Yes, I've created a page to list these kinds of points. For this one specifically I think it's only worth discussing once the embedding functionalities are added to see how it all makes sense together.

docs/contents/graphs.md
jpmoutinho: Changed the sentence to make it evident we show this later.

docs/contents/graphs.md
jpmoutinho: Not at all, I think that is a good idea. I've added it 👍

docs/contents/graphs.md
jpmoutinho: This is just a somewhat standard [adjacency matrix loading](https://networkx.org/documentation/stable/reference/generated/networkx.convert_matrix.from_numpy_array.html), with a slight modification of loading the diagonal $(i, i)$ as a weight on node $i$ instead of an edge $(i, i)$, because QoolQit assumes the graphs have no self-loops.

I made a slight change in the description to make this more explicit, and perhaps I can further clarify it depending on what you think is not clear.

We can further refine this later depending on what users need but I would still keep it for now, I don't think it's doing anything out of the ordinary and will be useful for testing how to structure the embedding part later.

docs/contents/graphs.md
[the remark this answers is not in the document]
jpmoutinho: Actually I take that back, I changed it to your suggestion. I ran into another thing where I ended up wanting to dispatch on the kwarg so I figured I could do it here as well.

docs/contents/graphs.md
nataliejpg: Okay I see, it wasn't clear before that if an off diagonal element was 0 no edge would be created. The extra documentation clears this up

nataliejpg (approved):
Looks good.

Issue #20, opened 2025-05-13
Design the `QuantumProgram`


jpmoutinho:
A `QuantumProgram` will bring together a sequence register and device, handle any potential unit conversion, potential sequence adjustment / compilation, and send for execution through Pulser. The design for this should take into account the ongoing work to refactor the Pulser backend API.

PR #35, opened 2025-06-17
Local backend integration

vytautas-a:
Integrate local Qutip and EMU-MPS backends with Qoolqit.

Closes https://github.com/pasqal-io/qoolqit/issues/20


qoolqit/execution/utils.py
```diff
@@ -9,3 +9,15 @@ class CompilerProfile(StrEnum):
     MAX_AMPLITUDE = "MaxAmplitude"
     MAX_DURATION = "MaxDuration"
     MIN_DISTANCE = "MinDistance"
+
+
+class BackendName(StrEnum):
+
+    QUTIP = "QutipBackendV2"
+    EMU_MPS = "EmuMPS"
```
jpmoutinho, line 17:
```suggestion
    QUTIP = "Qutip"
    EMUMPS = "EmuMPS"
```

qoolqit/program.py
```diff
@@ -26,10 +28,14 @@ def __init__(
         self,
         register: Register,
         drive: Drive,
+        backend_name: BackendName = BackendName.QUTIP,
+        **backend_params: Any,
```
jpmoutinho, line 32: Do you have a strong reason to pass the backend at program initialization? In principle it should be fine only at run-time?

jpmoutinho (commented):
Just a few quick comments

A design decision we should think about at this stage is if we want the user to do
```
program = QuantumProgram(..., backend_name = backend_name)
result = pogram.run()
```
or
```
program = QuantumProgram(...)
result = pogram.run(backend_name = backend_name)
```
or
```
backend = SomeBackend()
pogram = QuantumProgram(..., backend = backend)
result = program.run()
```
or
```
backend = SomeBackend()
pogram = QuantumProgram(...)
result = program.run(backend = backend)
```
or
```
backend = SomeBackend(...)
pogram = QuantumProgram(...)
result = backend.run(program)
```
I think I like the last one actually, but not sure. Do you think there are any potential implications to each one?

qoolqit/execution/backend.py
```diff
@@ -0,0 +1,124 @@
...
+import logging
+import string
+from collections import Counter
+from typing import Any, Union
+
+import emu_mps
+import numpy as np
+import pulser
+import torch
+from emu_mps import MPSBackend, MPSConfig
```
jpmoutinho, line 12: Don't forget to add emu_mps to the `pyproject.toml`

qoolqit/execution/backend.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def build_config(self, runs: int = 100) -> None:
+        # Add the necessary observables based on the expected result type
+        obs = self.backend_params.get("observables", [])
+
+        if len(obs) == 0:
+            if self.result_type == ResultType.BITSTRING:
+                obs.append(pulser.backend.BitStrings(num_shots=runs))
+            elif self.result_type == ResultType.STATE_VECTOR:
+                obs.append(pulser.backend.StateResult())
```
jpmoutinho, line 55: Are there more fields worth exposing from the Pulser backend configs? Check if it's worth creating a qoolqit backend config as some kind of simplified wrapper.

qoolqit/execution/backend.py
```diff
@@ -0,0 +1,124 @@
...
+
+class Backend:
+    """Wrapper for Pulser's backend API."""
+
+    def __init__(
+        self,
+        seq: Sequence,
+        name: BackendName = BackendName.QUTIP,
+        result_type: ResultType = ResultType.STATE_VECTOR,
+        **backend_params: Any,
```
jpmoutinho, line 34: What are you expecting in the backend_params?

qoolqit/execution/backend.py
```diff
@@ -0,0 +1,124 @@
...
+        self.backend_params.setdefault("log_level", logging.WARNING)
+
+        # Build the config object
+        self.config = self.config_cls(**self.backend_params)
+
+    def build_backend(self) -> None:
+        # Build the local backend
+        self.backend = self.backend_cls(self.seq, config=self.config)
+
+    def contract_mps(self, mps_state: emu_mps.MPS) -> torch.Tensor:
```
jpmoutinho, line 69: I think it would be nice to set the backend base class as something more abstract that does not have logic that is too specific to a single backend. Remember that this must work for all backends, including QPU. If this logic is temporary (because emu_mps is not yet integrated in Pulser), it would be best to keep these as some separately defined functions so we can keep this class definition as simple as possible.

qoolqit/execution/backend.py
```diff
@@ -0,0 +1,124 @@
...
+
+        einsum_str = ",".join(einsum_subs) + "->" + "".join(letters[n : 2 * n])
+        result = torch.einsum(einsum_str, *mps_state.factors)
+        return result.flatten().cpu()
+
+    def run(self, runs: int = 100) -> ResType:
+
+        # Build the config and the backend
+        self.build_config(runs)
+        self.build_backend()
```
jpmoutinho, line 98: Why is `build_backend` required at every run?

jpmoutinho (commented):
Thanks for the MR @vytautas-a. Let's converge a bit more on the structure and cleanup the code + add some tests for this MR, but I think it's going in the right direction.

qoolqit/execution/backend.py
jpmoutinho: Also, is emu_sv too different from emu_mps? Is it worth adding both in this MR?

vytautas-a: I can take a look at emu_sv as well. In principle it should follow the same new API, but I'm not familiar with it so there may be some caveats.

jpmoutinho (commented):
Thanks @vytautas-a, few minor comments.

qoolqit/program.py
```diff
@@ -118,16 +121,40 @@ def draw(
...
-    def run(self) -> ArrayLike:
-        """Temporary method to run a simulation on QuTip."""
+    def run(
+        self,
+        backend_name: BackendName = BackendName.QUTIP,
+        result_type: ResultType = ResultType.STATEVECTOR,
+        runs: int = 100,
+        evaluation_times: list[float] = [1.0],
+        **backend_params: Any,
+    ) -> OutputType:
```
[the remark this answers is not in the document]
vytautas-a: it seems that it's not necessary, `mypy` doesn't complain

qoolqit/execution/backend.py
vytautas-a: Track in #44 as well

jpmoutinho (approved):

PR #51, opened 2025-07-29
[Feature] Adding support for MyQLM as a backend

Yoric:
This is an early stage support.

To add support for method `proceed()`, we'll need a refactor.

Requesting review from @a-corni on MyQLM aspects, @vytautas-a on Qoolqit integration.

### DIFF
--- a/qoolqit/_solvers/backends/remote_backends.py
+++ b/qoolqit/_solvers/backends/remote_backends.py
@@ -5,12 +5,16 @@ import time
 from abc import abstractmethod
 from typing import Counter, cast
 
+import pulser
+import pulser.backend.remote
 from pasqal_cloud import SDK, EmulatorType
 from pasqal_cloud.batch import Batch
 from pulser import Sequence
-from pulser.backend.remote import BatchStatus
+from pulser.backend.remote import BatchStatus, RemoteResults
 from pulser.devices import Device
 from pulser.json.abstract_repr.deserializer import deserialize_device
+from pulser.result import SampledResult
+from pulser_myqlm import PulserQLMConnection
 
 from qoolqit._solvers.data import (
     BackendConfig,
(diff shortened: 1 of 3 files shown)


qoolqit/_solvers/backends/remote_backends.py
```diff
@@ -254,3 +258,134 @@ def _execute_remotely(self, sequence: Sequence, runs: int) -> JobId:
...
+        super().__init__(JobId(results.job_ids[0]))
+        self._remote_results = results
+        self._result: Result | None = None
+        self.sleep_duration_sec = sleep_duration_sec
+
+    def status(self) -> BatchStatus:
+        # As of this writing, we don't have a way to fetch the status.
+        if self._result is not None:
+            return BatchStatus.DONE
+        return BatchStatus.PENDING
```
a-corni, line 368: Is there an issue with self._remote_results.get_batch_status() ?

qoolqit/_solvers/backends/remote_backends.py
```diff
@@ -254,3 +258,134 @@ def _execute_remotely(self, sequence: Sequence, runs: int) -> JobId:
...
+
+    def wait(self) -> Result:
+        if self._result is not None:
+            return self._result
+
+        # Wait for execution to complete.
+        while True:
+            time.sleep(self.sleep_duration_sec)
+            results = self._remote_results.get_available_results()
+            if len(results) != 1:
```
a-corni, line 378: That's not exactly what you want: If you have submitted multiple Jobs in a batch, RemoteResults will be a dictionnary containing only the jobs that have finished.

I guest the best for you is to do `if self.id in results:`

qoolqit/_solvers/backends/remote_backends.py
```diff
@@ -254,3 +258,134 @@ def _execute_remotely(self, sequence: Sequence, runs: int) -> JobId:
...
+                self._status = BatchStatus.PENDING
+                continue
+
+            # From this point, we assume that len(expected) == 1, as this is the only
+            # way to construct a `RemoteQMLJob`.
+            result = results[self.id]
+            assert isinstance(result, SampledResult)
+
+            # FIXME: This is subject to race condition.
+            self._result = Result(counts=Counter(result.bitstring_counts))
```
a-corni, line 389: Why is it subject to race condition ?

a-corni (commented):
Thanks for your MR @Yoric !

I have some comments. On the whole, I have a suggestion regarding the API. To me, RemoteQPUBackend should be the entry point to any QPU, and you should specify in the BackendConfig on which remote ressource you want to look for it (you could have an enum to list these remote ressources, that could be "PasqalCloud" or "QLM", and we would map "PasqalCloud" to pulser_pasqal.PasqalCloud (instead of the SDK) and "QLM" to pulser_myqlm.PulserQLMConnection

qoolqit/_solvers/backends/remote_backends.py
```diff
@@ -254,3 +258,134 @@ def _execute_remotely(self, sequence: Sequence, runs: int) -> JobId:
...
+        if config.username is not None or config.password is not None:
+            logger.warning(
+                "ignoring username/password for the QLM backend "
+                + "in favor of external authentication"
+            )
+        self._connection = PulserQLMConnection()
+
+    def _api_max_runs(self) -> int:
+        # Arbitrary number, matching RemoteQPUBackend.
+        return 500
```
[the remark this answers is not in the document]
Yoric: It's a limitation of the server, not the QPU itself. As far as I can tell, there is no way to find out this magic value.

PR #52, opened 2025-08-05
new lattice types

### LINKED ISSUE #23: Add more graph constructors
Current graph constructors are basic. Quite a few are developed in qoolbox that should be migrated here.

- Geometric / unit disk graph constructors to the `DataGraph`
- Lattice graph constructors probably make more sense in the `LayoutGraph`

v-vitale:
Constructors for creating triangular, hexagonal and heavy hexagonal registers.


v-vitale:
Already taken into account in #60

PR #66, opened 2025-09-08
[Fix] Make it work under Windows

Yoric:
Windows compatibility was removed around version https://github.com/pasqal-io/qoolqit/commit/9bd63e84cff4a95a0095548f6ee20f4c1ad94ea3 . This breaks workflow for some of our users. We need it back.

### DIFF
--- a/docs/tutorials/basic_qubo.ipynb
+++ b/docs/tutorials/basic_qubo.ipynb
@@ -20,13 +20,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "f7a9b02b",
    "metadata": {},
    "outputs": [],
    "source": [
-    "from __future__ import annotations\n",
-    "\n",
     "import numpy as np\n",
     "\n",
     "Q = np.array([\n",
--- a/docs/tutorials/custom_quantum_program.ipynb
+++ b/docs/tutorials/custom_quantum_program.ipynb
@@ -30,13 +30,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 38,
+   "execution_count": null,
    "id": "e810874b",
    "metadata": {},
    "outputs": [],
    "source": [
-    "from __future__ import annotations\n",
-    "\n",
     "from qoolqit import DataGraph, Register\n",
     "\n",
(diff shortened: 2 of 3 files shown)


sgrava:
There is an MR ready in emu-mps to support windows: https://github.com/pasqal-io/emulators/pull/135
We could just wait for support from their side.

docs/tutorials/basic_qubo.ipynb
```diff
@@ -25,6 +25,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from __future__ import annotations\n",
+    "\n",
```
sgrava, line 29: Could you remove this from notebooks?

pyproject.toml
```diff
@@ -106,7 +106,7 @@ dependencies = [
   "mkdocstrings",
   "mkdocstrings-python",
   "mkdocs-section-index",
-  "mkdocs-jupyter==0.24.7",
+  "mkdocs-jupyter~=0.24",
```
sgrava, line 109: keep `mkodocs-jupyter `at max 0.24.7, since 0.25 won't render latex in notebooks correctly. A comment would help, my fault for not putting it.

.github/workflows/test.yml
```diff
@@ -44,3 +43,26 @@ jobs:
...
         if-no-files-found: ignore
+
+  test_qoolqit_windows:
+    # Windows requires its own syntax for testing Notebooks,
+    # to be implemented.
+    name: Run unit/integration tests (Windows)
+    strategy:
+      matrix:
+        os: [windows-latest]
+        python-version: ["3.13"]
```
chMoussa, line 54: Maybe at a later time, we can test other python versions.

chMoussa (approved):
LGTM @Yoric if you can apply @sgrava review.

.github/workflows/test.yml
Yoric: Yes, I only used 3.13 because tests are rather slow under Windows. Happy to put more versions.

docs/tutorials/basic_qubo.ipynb
Yoric: I have no idea how it got there. Removing.

PR #68, opened 2025-09-18
Device class refactoring for accepting a Pulser Device object

vytautas-a:
Currently to create a new `qoolqit` device class the user has to derive it from the base `Device` class and use the appropriate Pulser device in this derived class. This PR adds the capability for the user to simply pass any Pulser device to `qoolqit` `Device` constructor and get a working custom `qoolqit` device.

### DIFF
--- a/qoolqit/devices/device.py
+++ b/qoolqit/devices/device.py
@@ -146,9 +146,10 @@ class AnalogDevice(Device):
 
 class DigitalAnalogDevice(Device):
     """A device with digital and analog capabilites."""
+
     def __init__(self) -> None:
         super().__init__(pulser_device=_DigitalAnalogDevice)
-    
+
 
 class TestAnalogDevice(Device):
     def __init__(self) -> None:


qoolqit/devices/device.py
```diff
@@ -98,62 +112,48 @@ def set_distance_unit(self, distance: float) -> None:
...
-    def _device(self) -> BaseDevice:
-        return _TestAnalogDevice
-
-    @property
-    def _default_converter(self) -> UnitConverter:
-        return UnitConverter.from_energy(self._C6, self._upper_amp)
+        super().__init__(pulser_device=_TestAnalogDevice)
 
 
 ALL_DEVICES = [MockDevice, AnalogDevice, TestAnalogDevice, DigitalAnalogDevice]
```
v-vitale: Here ALL_DEVICES is a list hardcoded in qoolqit, ins't there anything in pulser.devices that is updated by pulser people?

v-vitale (changes requested):
I really like the idea of initializing the device as `Device(pulser_device = ...)`

1. I would change the value of `UPPER_DET.`
2. @sgrava we discussed with Francesco about showing the ranges of the control parameters I don't know whether we want to do it know or not
3. I am puzzled by the compiler when `UPPER_AMP` is set by hand meaning that one could call `CompilerProfile.MAX_AMP` and get a compilation with respect to an `UPPER_AMP` that is fictitious (e.g. in the `MockDevice` case). It would't throw any error but it is tricky. I see people using `MockDevice `only when they want to experiment, then we should probably also let them experiment with the values of `UPPER_DURATION`, _AMP and so on. They are instead hardcoded.

To make more explicit. Let's say someone want to do a time evolution with certain parameters on the mock device in such a way to check whether increasing the total time would change the final result.

The `UPPER_DURATION `of `MockDevice` will be hardcoded as 6000.
Now they can insert any value for the time since there are no limits in the device and the `QuantumProgram` would be compiled accordingly.
However if they select `CompilerProfile.MAX_DURATION` they would get a `Sequence` whose maximum duration is `UPPER_DURATION` that is smaller with respect to what they put as maximum time (since it was hardcoded)
It is not breaking anything but it seems confusing to me. 

All this to say that we can address it later but it is to be kept in mind.

PR #75, opened 2025-10-06
[Feature] Support detuning maps

Yoric:
A first draft of DMM support in `QuantumProgram`. Is this the kind of API we want to use?

Based on #49. However, this time, we're trying to provide an API that will be used not just by solvers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 name = "qoolqit"
 description = "A Python library for developing algorithms in the Rydberg Analog Model."
 readme = "README.md"
-version = "0.1.3"
+version = "0.2.0"
 requires-python = ">=3.10"
 license = { text = "MIT-derived" }
 keywords = ["quantum"]


tests/test_program.py
```diff
@@ -50,3 +51,25 @@ def test_compiler_profiles(
...
+@pytest.mark.parametrize("profile", CompilerProfile.list())
+def test_compiler_profiles_dmm(
+    device_class: Callable,
+    profile: CompilerProfile,
+    random_program_dmm: Callable[[], QuantumProgram],
+) -> None:
+
+    if profile != CompilerProfile.MIN_DISTANCE:
+        program = random_program_dmm()
+        print(program.register, program.drive.weighted_detunings)
```
sgrava: remove print

tests/conftest.py
```diff
@@ -75,6 +75,26 @@ def _generate_random_program() -> QuantumProgram:
...
+    random_linear_register: Callable[[], Register],
+    random_pos_ramp: Callable[[], Waveform],
+    random_neg_ramp: Callable[[], Waveform],
+) -> Generator[Callable[[], QuantumProgram]]:
+    def _generate_random_program() -> QuantumProgram:
+        register = random_linear_register()
+        wf_amp = random_pos_ramp()
+        wf_det = random_neg_ramp()
+        wdetuning = WeightedDetuning(
+            weights={q: uniform(0.1, 0.99) for q in register.qubits.keys()},
```
sgrava: nitpick
```suggestion
            weights={q: uniform(0.1, 0.99) for q in register.qubit_ids},
```

chMoussa: But I can add quickly.

PR #78, opened 2025-10-08
[Fix] We need specific versions of pasqal-cloud

Yoric:
See https://github.com/pasqal-io/maximum-independent-set/issues/141 .


sgrava (approved):
Nice!

Issue #85, opened 2025-10-14
Load Graphs methods in BaseGraph


v-vitale:
Add methods for loading from
- networkx object
- pyg object
and creating the relative BaseGraph

Steps: 
- Extract Nodes
- Extract Edges
- Check whether positions are present

PR #86, opened 2025-10-15
Importing graphs from networkx and pyg

v-vitale:
Connected to Issue #85.

## Changes
- `BaseGraph.from_pyg`
- `BaseGraph.from_nx`

Until we fully support input attributes as `np.array` and `torch`, the attributes of the graphs are heavily validated beforehand.

- tests

## Minor changes
- ENERGY factor is rounded to solve the rounding off error only for MAX_AMPLITUDE profile, It was causing tests to randomly fail. Full solution will be implemented in the future for the issue #102

### DIFF
--- a/qoolqit/execution/compilation_functions.py
+++ b/qoolqit/execution/compilation_functions.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import math
+
 from pulser.devices import Device as PulserDevice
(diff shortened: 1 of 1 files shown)


v-vitale:
To keep in mind that we should define a type that includes Float,Int,Real,torch.Tensor for nodes/weights/pos to make the framework more generalizable

PR #90, opened 2025-10-17
fix counter cast in _solvers

### LINKED ISSUE #87: Counter results is not turned into a Counter as expected
The casting of the result returned by remote backend from a dict to a `Counter` is not working as expected [here](https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/_solvers/backends/remote_backends.py#L97). 

Quick example to reproduce:

```python
>>> from typing import Counter, cast
>>> my_dict = {}
>>> counter = cast(my_dict,Counter)
>>> isinstance(counter, Counter)
False
```

Instead you can easily turn the dict into a counter as such `Counter(my_dict)`.

This leads to an exception being raised in the MIS library [here](https://github.com/pasqal-io/maximum-independent-set/blob/main/mis/solver/solver.py#L274) as the assertion fails when using remote backends.

sgrava:
Resolve #87

### DIFF
--- a/qoolqit/_solvers/backends/remote_backends.py
+++ b/qoolqit/_solvers/backends/remote_backends.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 import logging
 import time
 from abc import abstractmethod
-from typing import Counter, cast
+from collections import Counter
+from typing import cast
 
 from pasqal_cloud import SDK, EmulatorType
 from pasqal_cloud.batch import Batch
@@ -85,7 +86,6 @@ class RemoteJob(BaseJob):
             batch.refresh()
             if batch.status in {"PENDING", "RUNNING"}:
                 # Continue waiting.
-
                 continue
             job = batch.ordered_jobs[0]
             if job.status == "ERROR":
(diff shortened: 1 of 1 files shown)


sgrava:
Hi @MatthieuMoreau0, 
is there an example I can give it a try on for this fix?

I'm running this [notebook](https://github.com/pasqal-io/maximum-independent-set/blob/main/examples/tutorial%201%20-%20Using%20a%20Quantum%20Device%20to%20solve%20MIS.ipynb).
The notebook works fine with all remote backends, even with the QPU (suspiciously), but I'm not able to run any job run on my cloud portal.

MatthieuMoreau0:
In that notebook the pre-processing step solves the MIS without any quantum simulation.. Change the graph to study with the path `a265032_1dc.64.txt` and the issue should pop up

sgrava:
> In that notebook the pre-processing step solves the MIS without any quantum simulation.. Change the graph to study with the path `a265032_1dc.64.txt` and the issue should pop up

Indeed, changing the graph/path to study trigger the issue.
Moving to this branch indeed solves it.

PR #91, opened 2025-10-20
Sg/83 make backend choice easier for users and document

sgrava:
### Changes
- [x] rework the notebooks with the new backends
- [x] update the `docs/content/execution.md ` page


docs/contents/execution_deprecated.md
```diff
@@ -0,0 +1,98 @@
+This page refers to a soon to be deprecated way to run quantum programs as `program.run(backend_type=...)`.
```
v-vitale: Can we say since which version of qoolqit it will be considered deprecated? Such that the users can now if it is in their case.

v-vitale (approved):
It looks good to me,  it is very clear and I like the difference between submit and run for RemoteEmulators.

Just a small suggestion to pin the version after which the previous way to submit job is considered deprecated.

PR #101, opened 2025-10-27
add _to_pulser method

### LINKED ISSUE #100: WaveformConverter that does not rely on CustomWaveform
Add a symbolic translation layer between `qoolqit` and `pulser` Waveforms.

## Description
Currently, the conversion from `qoolqit.Drive` to `pulser.Pulse` at the compilation stage, is done by:
1.  sampling the Drive waveforms with 1 ns step
2. creating `pulser.CustomWaveform`s for amplidude/detuning from samples
3. output a Pulse built from the waveforms created in 2.
4. add this pulse to the output `pulser.Sequence` of the compilation stage

In practice, this approach makes the sequence a very big object (size ~MB) because it has to store **all** the values and ns resolution.
As a consequence, cloud struggle to handle the abstract representation of the sequence for remote execution.

## Actions
- [ ] each waveform must have a `_to_pulser` method that implements the conversion to a `pulser.Waveform`
- [ ] rewrite `WaveformConverted` to handle the translation symbolically and not by values

sgrava:
Resolve #100

## Changes
- removing default conversion through CustomWaveform.
- adding a `_to_pulser` method on each waveform to handle the conversion.
- if a waveform does not implement `_to_pulser` this defaults now to the interpolated waveform.
- `_to_pulser` on composite waveform is tricky because pulser round all durations of the waveforms. For this reason, `round_to_sum` is introduced.

### DIFF
--- a/qoolqit/program.py
+++ b/qoolqit/program.py
@@ -1,23 +1,22 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Union
 from warnings import warn
 
 import matplotlib.pyplot as plt
+from matplotlib.figure import Figure
 from pulser.sequence.sequence import Sequence as PulserSequence
 
 from qoolqit.devices import Device, MockDevice
 from qoolqit.drive import Drive
-from qoolqit.execution.backend import BaseBackend, EmuMPSBackend, OutputType, QutipBackend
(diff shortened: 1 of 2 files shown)


vytautas-a (approved):
Looks good

PR #104, opened 2025-11-05
add interpolated waveform

### LINKED ISSUE #103: add interpolated waveform
Add to QoolQit waveforms an interpolated waveform with its `_to_pulser` method for translation.

sgrava:
Resolve #103

### DIFF
--- a/tests/test_waveforms.py
+++ b/tests/test_waveforms.py
@@ -162,6 +162,14 @@ def test_interpolated_wrong_times_len() -> None:
         Interpolated(10, values=[0, 1], times=[0, 0.5, 0.8])
 
 
+def test_interpolated_to_pulser() -> None:
+    values = [0.1, 0.3, -0.5, 1.0, 5.7]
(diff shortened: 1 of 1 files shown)


qoolqit/waveforms/waveforms.py
```diff
@@ -128,6 +132,70 @@ def __repr_header__(self) -> str:
...
+        return float(self._values.min())
+
+    def max(self) -> float:
+        return float(self._values.max())
+
+    def _to_pulser(self, duration: int) -> ParamObj | pulser.InterpolatedWaveform:
+        return pulser.InterpolatedWaveform(
+            self.duration,
+            values=self._values,
+            times=self._times,
```
chMoussa: @sgrava I think the passed values and times should be also added as arguments to `_to_pulser`, because they need conversion similarly to the duration.

chMoussa: Ah nvmd, I guess because we use fractional values, we should be good to go.

PR #135, opened 2025-12-04
remove list of available devices

### LINKED ISSUE #134: device supported profiles
Remove default values, non-restricted device specs.
Let the device chose its supported compilation profiles.

## Description
QoolQit keeps a list of supported devices [[here](https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/devices/device.py)](https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/devices/device.py)

```python
ALL_DEVICES = [MockDevice, AnalogDevice, TestAnalogDevice, DigitalAnalogDevice]
```

and defines supported compilation profiles only for devices in this list.

  This is bad for several reasons:

- does not allow defining new devices, either user defined or remote ones (Fresnel, Fresnel_Can1, R&D devices, other 30 new devices by 2030)
- even letting the user defining new devices, compilation profiles will not be supported because this new device is not in the available device list

### Solution

1. Let each device decide which compilation profile to support. Something like:

```python
class MockDevice(Device):
			max_amp = None
			supported_profiles = (basic_compilation,)

class FresnelDevice(Device):
			max_amp = None
			supported_profiles = (basic_compilation, max_duration, max_amplitude,..)
```

For example, MockDevice should just have the basic compilation profile because it has no bounds. The strategy MAX_DURATION, for example, on a MockDevice really does not make sense as explained by @Vittorio Vitale in this [[issue](https://github.com/pasqal-io/qoolqit/issues/89)](https://github.com/pasqal-io/qoolqit/issues/89)

2. As a bonus point, one can get rid of default values set [[here](https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/devices/device.py#L11)](https://github.com/pasqal-io/qoolqit/blob/main/qoolqit/devices/device.py#L11)

UPPER_DURATION = 6000
UPPER_AMP = 4.0 * pi
UPPER_DET = 4.0 * pi
LOWER_DISTANCE = 5.0

sgrava:
Resolve #134 

allow the user to define new devices from pulser remote devices. For now as
```python
my_remote_pulser_device = connection.fetch _available_devices()["FRESNEL"]
qoolqit_device = Device(pulser_device=my_remote_pulser_device)
```

### Changes
- remove list of available devices.
- apart from default devices, mock, analog and digitalanalog, all new devices will be created from a pulser device.
- docs and docstrings improvements
- device.info() shows a short description and specs.
```python
>>> AnalogDevice().info()
AnalogDevice: A realistic device for analog sequence execution.
 └── max_duration: 75.39822368615503
 └── max_amplitude: 1.0
 └── max_detuning: 10.0
 └── min_distance: 0.7809234915702248
```
- qoolqit.available_default_devices() shows all default devices

### Extra
- [x] Moved the discussion about creating a device from a connection here #137

### DIFF
--- a/tests/test_program.py
+++ b/tests/test_program.py
@@ -6,8 +6,8 @@ from typing import Callable
 import pytest
 from pulser.sequence import Sequence as PulserSequence
 
-from qoolqit import __version__ as qoolqit_version
 from qoolqit import AnalogDevice, DigitalAnalogDevice, MockDevice
+from qoolqit import __version__ as qoolqit_version
 from qoolqit.drive import Drive
 from qoolqit.exceptions import CompilationError
 from qoolqit.execution import CompilerProfile


a-quelle-pasqal (changes requested):
Might I interest you in also extending `QuantumProgram.compile_to` to accept a `pulser.Device`, and then simply construct a `Device(device)` with the default unit converter in that function if a `pulser.Device` was given. The change is minor, would make the now available workflow slightly easier, and would allow you to delete all the `Device` subclasses, which don't really add anything new.

On the other hand, this MR is already a good improvement on its own, and if you choose to ignore it, I won't object.

sgrava:
> Might I interest you in also extending `QuantumProgram.compile_to` to accept a `pulser.Device`, and then simply construct a `Device(device)` with the default unit converter in that function if a `pulser.Device` was given. The change is minor, would make the now available workflow slightly easier, and would allow you to delete all the `Device` subclasses, which don't really add anything new.
> 
> On the other hand, this MR is already a good improvement on its own, and if you choose to ignore it, I won't object.

@a-quelle-pasqal 
Unfortunately, the device is not a simple copy of the pulser device, but also hosts the unit converter, which bridge the gap between a QooQit quantum program in dimensionless units and a pulser sequence in the units of the device.

We really want user to use QooQit devices. I can decide here to show only relevant stuff to the user, like only the specs he/she needs.

a-quelle-pasqal (approved):
LGTM

Issue #153, opened 2026-01-16
Add blade embedder


sgrava:
Add the blade embedding algorithm from[ qubo-solver](https://github.com/pasqal-io/qubo-solver) library.

## Description
To reuse the same embedding algorithm in other libraries (qubo, mis, qek) it is convenient to refactor such code into a QoolQit embedder.

## Actions
- [x] organize `qoolqit/embedding/` and `tests/test_embedding` to prepare for blade embedding porting
      - issue: #155
      - PR: #156  
- [x] port the algorithm in `qoolqit/embedding/algorithm`: done in #154 
     - [x] investigate if `scikit-learn` and `seaborn` are necessary as dependencies: optional
     - [x] test
     - [x] add a `BladeEmbedder` and a `BladeEmbedderConfig` with optional device dependency.
     - [x] integration test 
     - PR: #154 
- [ ] expand documentation to add `BladeEmbedder`
     - issue: #161 
     - PR: 
- [ ] Refactor `blade_embedding/_qubo_mapper.py`
     - issue: #162 
- [ ] add blade integration test #165 
- [ ] blade for device: how to see the final pulser register #168

PR #154, opened 2026-01-16
Add blade embedder from qubo-solver

sgrava:
Related to #153

Port [qubo-solver/algorithms/blade](https://github.com/pasqal-io/qubo-solver/tree/main/qubosolver/algorithms/blade) such that all application library can use it without depending on `qubo-solver`.

## Usage
```python
import numpy as np
from qoolqit import BladeEmbedder, BladeEmbeddingConfig

# random qubo
qubo = np.random.random((5,5))
qubo += qubo.mT
qubo /= qubo.max()

# instantiate blade embedder, optional config
config = BladeEmbeddingConfig(max_min_dist_ratio=7.0)
embedder = BladeEmbedder(config=config)

graph = embedder.embed(qubo)
```

#### With Device info
Make a custom config directly from the device. It basically just overwrite `max_min_dist_ratio`.
```python
from qoolqit import AnalogDevice

config = BladeEmbeddingConfig(device=AnalogDevice())
embedder = BladeEmbedder(config=config)
print(embedder)
```
```sh
BladeEmbedder:
| Algorithm: blade_embedding
| Config: BladeEmbeddingConfig(max_min_dist_ratio=7.6, dimensions=(5, 4, 3, 2, 2, 2), starting_positions=None, pca=False, steps_per_round=200, compute_weight_relative_threshold=<function BladeEmbeddingConfig.<lambda> at 0x71a450b9e8e0>, compute_max_distance_to_walk=<function BladeEmbeddingConfig.<lambda> at 0x71a450b9e980>, starting_ratio_factor=2, draw_steps=False)
```

#### Porting
- add `qubo-solver/algorithms/blade` as embedder into `qoolqit/embedding/algorithms/blade`.
- add tests from [qubo-solver/tests/embeddings/test_blade.py](https://github.com/pasqal-io/qubo-solver/blob/main/tests/embeddings/test_blade.py)
- removing `em_blade_for_device` as we offer that option from the config.
- lint, format and fix some typos.

#### Comply with BaseEmbedder
- add `BladeEmbedder` and `BladeEmbeddingConfig` classes
- add `device` kwarg in `BladeEmbeddingConfig`, which overwrite the config to fix blade constrains to device specs

## To discuss
- [x] documentation in a separate issue/MR #161 
- [x] scikit-learn dependency.
      The algorithm depends on sklearn only because of `sklearn.decomposition.PCA` which can be replaced by an equivalent (up to eigenvectors sign) truncated `numpy.linalg.svd`. 
      For now, let's keep everything as an optional dependency.
- [x] Is all the drawing turned on by the `draw_step` kwarg necessary? Let's keep it for now.
- [x] running the blade algorithm it is common to get a warning. I propose to fix it separately in a different issue #163.
```sh
 /home/stefano/Workspace/qoolqit/qoolqit/embedding/algorithms/blade_embedding/_helpers.py:9: RuntimeWarning: divide by zero encountered in divide
  return 1 / dist**6
/home/stefano/Workspace/qoolqit/qoolqit/embedding/algorithms/blade_embedding/_helpers.py:13: RuntimeWarning: divide by zero encountered in divide
  return (1 / weight) ** (1 / 6)  # type: ignore[no-any-return]
```


qoolqit/embedding/algorithms/blade_embedding/_dimension_shrinker.py
```diff
@@ -0,0 +1,65 @@
...
+    assert (space_dimension,) == dimensions_lengths.shape
+
+    minima_per_dim = np.min(positions, axis=0)
+    maxima_per_dim = np.max(positions, axis=0)
+    current_lengths = maxima_per_dim - minima_per_dim
+    lengths_to_remove = np.maximum(current_lengths - dimensions_lengths, 0)
+
+    new_minima = minima_per_dim + lengths_to_remove / 2
+    new_maxima = maxima_per_dim - lengths_to_remove / 2
+    shrinked_positions = np.maximum(new_minima, np.minimum(new_maxima, positions))
```
abussy-pasqal: Use `np.clip` ?
https://numpy.org/doc/2.0/reference/generated/numpy.clip.html

qoolqit/embedding/algorithms/blade_embedding/_qubo_mapper.py
```diff
@@ -0,0 +1,299 @@
...
+        a = [[1, -2, -3], [0, -1, -1], [0, 0, -3]]
+
+        corresponds to the following Qubo:
+
+        Q(x) = x0 -2*x0x1 -x1 -3*x0x2 -x1x2 -3*x2
+
+        Notice that the resulting matrix is always upper triangular.
+
+        Args:
+            variables: If present, the returned matrix will follow the same
```
abussy-pasqal: I'm not sure to understand what `variables` is. Maybe you could add a small unit test, that could also be an example ?
Also, can an index be present multiple time in variables ?

tests/test_embedding/test_blade_embedding.py
```diff
@@ -0,0 +1,232 @@
...
+    qubo_graph.add_edge(0, 1, weight=normalized_interaction(1))
+
+    new_positions = update_positions(
+        positions=np.array([[-1, 0], [1, 0]]),
+        qubo_graph=qubo_graph,
+        min_dist=10,
+        max_distance_to_walk=(0, 2, 0),
+    )
+
+    assert (new_positions == np.array([[-3, 0], [3, 0]])).all()
```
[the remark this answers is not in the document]
sgrava: Good point!
Done

qoolqit/embedding/algorithms/blade_embedding/_distances_constraints_calculator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    differences = embedded_qubo_triu - target_qubo_triu
+
+    percent = 100 - 2 / (len(target_qubo) - 1) * 10
+    percentile = np.percentile(differences, percent)
+
+    difference_ceiling = max(0.0, percentile)  # type: ignore
+    limited_differences = np.minimum(differences, difference_ceiling)
+
+    if plot:
```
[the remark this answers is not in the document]
sgrava: Done! 
Now `seaborn` is lazily imported in the only drawing function that requires it.

> It also separates plotting and algorithms

That is the direction I agree, but I cannot really do all of that here in one time.

qoolqit/embedding/matrix_embedder.py
```diff
@@ -38,3 +43,29 @@ class InteractionEmbedder(MatrixToGraphEmbedder[InteractionEmbeddingConfig]):
...
+        """Return a DataGraph with coordinates that embeds the input matrix.
+
+        Validates the input, runs the embedding algorithm, and validates the output.
+
+        Args:
+            data (np.ndarray): the matrix to embed into a DataGraph with coordinates.
+        """
+        self.validate_input(data)
+        positions = self.algorithm(data, **self.config.dict())
+        centered_coords = positions - np.mean(positions, axis=0)
```
CdeTerra: The positions should not be centered by subtracting the mean coordinates because BLaDE already returns positions relative to a center with null coordinates, and the mean of the positions is not necessarily the best center.

tests/test_embedding/test_blade_embedding.py
```diff
@@ -0,0 +1,243 @@
...
+        qubo_graph=qubo_graph,
+        max_dist=1,
+        max_distance_to_walk=(0, 0, 1),
+        draw_step=True,
+    )
+    assert len(plt.get_fignums()) > 0
+    plt.close("all")
+
+
+def test_with_device() -> None:
```
CdeTerra: I initially wrote this test case to check the `em_blade_for_device` function. We need to confirm the function should be removed to adapt this test without the device (in which case the UT should be renamed without `device`).

Also, I've found in qoolbox's code tests involving the C6 coefficient based on Pulser's `interaction_coeff`, but I haven't found tests that check how a specific register is adapted to this value.

sgrava: See my comment above, let's fix it in another issue #165

Issue #152, opened 2026-01-16
`from_pyg` method is too restrictive


ShaheenPasqal:
The `BaseGraph` class have a method to convert a pytorch geometric graph into QoolQit graph.

However, there are some assumptions about the node and edge features in the input. 

Indeed, [it is expected that the tensors x and edge_attr have a specific shape](https://github.com/pasqal-io/qoolqit/blob/050a442772c0e72b207dc510b318c7e68e6fba45/qoolqit/graphs/base_graph.py#L204C1-L209C10) (`[number of nodes, 1]` for x and `[number of edge, 1]` for the edge), but there is no restriction on their shapes in torch geometric datasets. 

For instance, in the [PTC-FM](https://pytorch-geometric.readthedocs.io/en/2.7.0/generated/torch_geometric.datasets.TUDataset.html) dataset used for the QEK algorithm, the torch geometric features shape are [number of node, 18] and [number of edges, 4], respectively for `x`and `edge_attr`. 

## Proposal:

Do not use `x`and `edge_attr` to construct the `BaseGraph`QoolQit object.

PR #170, opened 2026-02-05
Refactor from_pyg and add to_pyg methods to BaseGraph

MGarrigues-pasqal:
## Motivation

#152 : The previous `from_pyg` implementation was rigid: it enforced fixed tensor shapes (e.g., x had to be (n,1)) and didn't easily allow choosing which attributes to keep.

This PR makes the PyG <-> QoolQit conversion more flexible by relying on PyG's native utilities (`to_networkx`, `from_networkx`). Standard PyG attributes (x, edge_attr, y, pos) and QoolQit attributes (weight, edge_weight) are automatically transferred. Users can specify additional attributes via `node_attrs`, `edge_attrs`, and `graph_attrs` parameters.

I tried to keep as much as possible to the way of doing things in networkx with every data being either node, edge or graph attribute, as lists. What do you think, does it align with what you had in mind ?

Same goes for to_pyg, I added it as I am using in harmoniqa, but you may not wish to include it.

## Changes

- **`from_pyg`**: uses PyG's `to_networkx`, automatically includes standard PyG attrs (x, edge_attr, y, pos) + QoolQit attrs (weight, edge_weight)
- **`to_pyg`** (new): uses PyG's `from_networkx`, preserves tensor types on roundtrip
- **QoolQit attributes**: weight, edge_weight are mapped to internal dicts (_node_weights, _edge_weights); pos is mapped to _coords
- **Refactoring**: `_reset_dicts`, `has_node_weights`, `has_edge_weights` moved from DataGraph to BaseGraph, since BaseGraph instances can have weights when created via `from_nx` or `from_pyg`

## Test plan
- [x] Existing tests adapted and passing (36 tests)
- [x] New roundtrip tests for QoolQit attrs (weight, pos, edge_weight, y)
- [x] New roundtrip tests for PyG attrs (x, edge_attr)
- [x] Pre-commit checks passing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,9 @@ authors = [
   { name = "Manu Lahariya", email = "manu.lahariya@pasqal.com" },
   { name = "Charles MOUSSA", email = "charles.moussa@pasqal.com" },
   { name = "Vittorio Vitale", email = "vittorio.vitale@pasqal.com" },
-  { name = "Clément de Terrasson de Montleau", email = "clement.de-terrasson@pasqal.com" }
+  { name = "Clément de Terrasson de Montleau", email = "clement.de-terrasson@pasqal.com" },
+  { name = "Mathieu Garrigues", email = "mathieu.garrigues@pasqal.com" }
+
 ]
 classifiers = [
   "Development Status :: 3 - Alpha",


MGarrigues-pasqal:
Thank you @sgrava, yes this looks better I will change it !

As for the edge and node weights, what do you think of the following approach, trying to keep the best of both worlds. Adding "node_weights_attr" and "edge_weights_attr" as str arguments in from_pyg function (and keeping the other args for edge, node and graph attrs). This way we have a "clean" and coercive way to assign weights (I'll add dimension and type checks), while keeping the flexibility of adding other attributes for the user. What do you think ?

sgrava (approved):
Looking good! Many thanks for the contribution

PR #176, opened 2026-02-16
Adapt dimensions for starting positions in BLaDE + rename 'qubo' into 'interactions'

CdeTerra:
Migrated from https://github.com/pasqal-io/qubo-solver/pull/95

### DIFF
--- a/tests/test_embedding/test_blade.py
+++ b/tests/test_embedding/test_blade.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import networkx as nx
 import numpy as np
 import pytest
+import scipy
 
 from qoolqit import AnalogDevice, BladeConfig, DigitalAnalogDevice, MockDevice
 from qoolqit.devices import Device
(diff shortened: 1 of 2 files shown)


qoolqit/embedding/algorithms/blade/blade.py
sgrava: Just a suggestion, but I think it is preferable to just pass a matrix rather than a graph on core algorithm parts that compute something.
It makes it better for testing, it matches other embedders, and you don't have to rely on an additional object. Just a numpy array. 

Also, I would move `draw_steps=draw_steps,` argument at the bottom.
The idea is to have first the important parameters and then the log/draw things which are optional.

tests/test_embedding/test_blade.py
```diff
@@ -183,21 +183,37 @@ def test_high_dimension_increase_after_equilibrium() -> None:
...
+            [0, 0, 1, 0],
+            [0, 0, 0, 1],
+            [0, 0, 0, 0],
+        ]
+    )
+
+    positions = blade(qubo)
+    positions = blade(qubo, starting_positions=positions)
+
+    assert np.all(positions)
```
sgrava: Please, clarify this test.
Ideally, I would like to see something like:

```python
# no starting positions
positions_no_sp = blade(qubo)
expected_positions_no_sp = ???
np.testing.assert_allclose(positions_no_sp, expected_positions_no_sp)

# with starting positions
positions = blade(qubo,  starting_positions=positions)
expected_positions = ???
np.testing.assert_allclose(positions, expected_positions)
```

sgrava:
Hi @CdeTerra thanks!
Looks really nice, just few minors really.

qoolqit/embedding/algorithms/blade/blade.py
CdeTerra: Yes, I agree, I'd like to do this small refactor in the future and consistently rely on the matrix. I also agree on the parameter point, although I'll probably also make many of them keyword-only parameters.

tests/test_embedding/test_blade.py
CdeTerra: I've made the test more explicit and added distances checking.

qoolqit/embedding/algorithms/blade/blade.py
CdeTerra: Done in https://github.com/pasqal-io/qoolqit/pull/178

PR #192, opened 2026-03-04
Fix typing in interaction_embedding

### LINKED ISSUE #191: Wrong typing in interaction_embedding
```python
def interaction_embedding(matrix: np.ndarray, method: str, maxiter: int, tol: float) -> np.ndarray:
```
From the code, it should return a `DataGraph`.

```python
 def cost_function(new_coords: np.ndarray, matrix: np.ndarray) -> np.float:
```
`np.float` doesn't exist. It should be `np.floating[Any]` instead.

abussy-pasqal:
Fixes #191


sgrava (approved):
Looks good, thanks!

Issue #177, opened 2026-02-17
Investigate random blade position scaling


sgrava:
Investigate why, randomly, the scaling of the positions returned by blade looks off.

## Description
A good test for blade would be to return expected positions.
To do so, we generate simple positions, to get the corresponding interaction matrix.
Then try to reverse engineer them with the embedder.

The issue is that depending on the seed, this simple test fails for blade.

## Acceptance criteria
- [ ] Identify the cause
- [ ] decide if to solve it here or in another issue

## To reproduce
On python 3.13

```python
import numpy as np
from qoolqit.embedding.algorithms.blade.blade import blade

def interaction_matrix(X: np.ndarray) -> np.ndarray:
    n = len(X)
    J = np.zeros((n, n))
    for i in range(n):
        for j in range(i + 1, n):
            J[i, j] = J[j, i] = 1 / np.linalg.norm(X[i] - X[j]) ** 6
    return J

target_positions = np.array([[-1.5,0.0], [-0.5,0.0], [6.0,0.0]], dtype=np.float64)
target_interactions = interaction_matrix(target_positions)

np.random.seed(4)
blade_pos = blade(target_interactions)
blade_interactions = interaction_matrix(blade_pos)
stress_blade = np.linalg.norm(target_interactions-blade_interactions)
np.testing.assert_allclose(stress_blade,0.0, atol=1e-2)
```

PR #212, opened 2026-03-20
Add feature to regulate priorities + fix filter differences for scaling...

CdeTerra:
- Add a new feature to regulate the priorities on the forces. It is configurable step by step and set to a default value. It allows to smoothen the priorities to avoid freezing situations, and avoids the situation of completely ignoring responsible objectives.
- Fix filter differences for scaling that could lead to a wrong scaling and situations where the forces had constantly to compensate.
- Make the step index global
- Add many tests (some of which quite tricky) including the macroscopic & microscopic testing for #177 
- Improve a lot the plotting
- Other fixes, enhancements and cleaning

This replaces #184's method that couldn't both fix the instability of positions and the constraint enforcement

Example of plotting:

image

### DIFF
--- a/qoolqit/embedding/algorithms/blade/_distances_constraints_calculator.py
+++ b/qoolqit/embedding/algorithms/blade/_distances_constraints_calculator.py
@@ -15,9 +15,10 @@ def compute_best_scaling_for_qubo(
     filter_differences: bool = True,
     draw_differences: bool = False,
 ) -> Any:
-    """Computes the scaling factor on the positions to minimize the distance.
+    """Computes the best scaling factor on the positions.
 
-    between the target and the embedded interactions.
+    The best scaling factor minimizes the distance between the target and
+    the embedded interactions.
     """
 
     embedded_interactions_triu = embedded_interactions[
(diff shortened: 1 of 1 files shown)


CdeTerra:
Fixes #177

tests/test_embedding/blade/test_blade.py
```diff
@@ -0,0 +1,530 @@
...
+    np.fill_diagonal(target_interactions, 0)
+
+    for seed in range(15):
+        np.random.seed(seed)
+        blade_pos = blade(
+            target_interactions, dimensions=dimensions, steps_per_round=steps_per_round
+        )
+        blade_interactions = interaction_matrix_from_positions(blade_pos)
+        quality = np.linalg.norm(target_interactions - blade_interactions)
+        np.testing.assert_allclose(quality, 0.0, atol=1e-2)
```
sgrava: I assume the fix solved the scaling issue. 
We can keep the test if you feel it but please, remove the seed loop.
It was useful to pin the issue, but now has lost specificity.

sgrava:
Hi @CdeTerra, looking good overall! Thanks!
Just a few minors.

Issue #84, opened 2025-10-13
track emulators issues


sgrava:
Track the following issues in emulators until they are resolved
- https://github.com/pasqal-io/emulators/issues/169
- https://github.com/pasqal-io/emulators/issues/168

PR #220, opened 2026-03-27
bump pulser and emulator

### LINKED ISSUE #219: bump pulser
bump pulser to 1.7

sgrava:
Resolve #219 and #84

## Changes
- getting rid of `job_params` in backends
- runs does not have a default value any more
     - for emulators the default is 1000, and it is cooked inside the emulation config
     - for QPU it must be specified
- using new pulser functionalities to simplify the logic. `EmulationConfig` can be updated with the method `with_changes` so it is easy now to add bit strings if not present.
- adding some QPU initialization tests that were missing

## To do
- runs -> num_shots to comply with pulser new naming

### DIFF
--- a/qoolqit/execution/backends.py
+++ b/qoolqit/execution/backends.py
@@ -252,7 +252,7 @@ class QPU(PulserRemoteBackend):
             raise ValueError(
                 """Number of runs must be provided to use the QPU backend.
                 Please specify the number of runs when creating the QPU instance.
-                For example: QPU(connection=..., runs=100""",
+                For example: QPU(connection=..., runs=100)""",
             )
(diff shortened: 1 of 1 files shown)


qoolqit/execution/backends.py
```diff
@@ -219,60 +198,97 @@ def run(self, program: QuantumProgram) -> Sequence[Results]:
...
+            username="your_username",
+            password="your_password",
+            project_id="your_project_id"
+        )
         backend = QPU(connection=connection)
         remote_results = backend.submit(program)
         ```
-    """  # noqa
+
+        Using Atos MyQML:
```
v-vitale: Have we checked whether it works for sending jobs to Julich and CEA?

PR #239, opened 2026-04-03
Doc v1.0

v-vitale:
I have an issue with list of contents not appearing on the right columns @sgrava . Do you know whether there's something in particular to do when writing in Markdown?


sgrava (approved):
Nice!

PR #261, opened 2026-04-13
add failing tests that should pass

### LINKED ISSUE #259: Bug in compilation with device_max_duration_ratio
device_max_duration_ratio rescaling after program validation is performed so it only works when the inputted duration is less than the maximum

Minimal working example:

```
from qoolqit import DataGraph,Register
from qoolqit.waveforms import Constant
from qoolqit import AnalogDevice


reg=Register.from_graph(DataGraph.line(2))
amp=Constant(1000,1)
drive=Drive(amplitude=amp)
program=QuantumProgram(reg,drive)
program.compile_to(AnalogDevice(),device_max_duration_ratio=1.0)
```

sgrava:
Resolves #259 

The bug originates in compilation.
The duration of a program is checked before the flag `device_max_duration_ratio` can manipulate the program, thus the compilation error.

## To solve
To solve at the moment produce a bit of code duplication:
- conditional check in `_validate_program_max_energy`
- conditional check in `_validate_program_default`

In both instances the duration check is disabled if
- device does not have a max duration
- `device_max_duration_ratio` is selected

## Changes
- add a failing test. As suggested in #259, set large duration and `device_max_duration_ratio=1.0` together.
- added logic to disable duration check is `device_max_duration_ratio` is selected

### DIFF
--- a/qoolqit/execution/compilation_functions.py
+++ b/qoolqit/execution/compilation_functions.py
@@ -93,7 +93,7 @@ def basic_compilation(
     """
     if profile == CompilerProfile.WORKING_POINT:
         TIME, ENERGY, DISTANCE = device.converter.factors
-        _validate_program_default_profile(register, drive, device)
+        _validate_program_default_profile(register, drive, device, device_max_duration_ratio)
     elif profile == CompilerProfile.MAX_ENERGY:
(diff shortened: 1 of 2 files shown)


v-vitale (approved):
LGTM

PR #262, opened 2026-04-13
allowing to specify xo in interaction embedder

v-vitale:
As per description, the `InteractionEmbedderConfig` was always starting from a random vector `x0`. 

Here we allow the selection of `x0` by means of the `InteractionEmbedderConfig`, defaults to `None` otherwise. 

When `None` it initializes a random `x0` vector as before.


tests/test_embedding/test_interaction_embedding.py
```diff
@@ -40,3 +40,14 @@ def test_interaction_embedding(n_qubits: int) -> None:
...
+    x0 = np.random.rand(n_qubits * 2)
+    embedder.config.x0 = x0
+    embedded_graph_x0_1 = embedder.embed(matrix)
+    coords_x0_1 = np.array(list(embedded_graph_x0_1.coords.values()))
+
+    embedded_graph_x0_2 = embedder.embed(matrix)
+    coords_x0_2 = np.array(list(embedded_graph_x0_2.coords.values()))
+
+    # Results should be identical (up to numerical precision)
+    assert np.allclose(coords_x0_1, coords_x0_2, atol=1e-6)
```
sgrava:
```suggestion
    np.testing.assert_allclose(coords_x0_1, coords_x0_2, atol=1e-6)
```

qoolqit/embedding/algorithms/interaction_embedding.py
```diff
@@ -47,7 +50,8 @@ def cost_function(new_coords: np.ndarray, matrix: np.ndarray) -> np.floating[Any
     np.random.seed(0)
 
     # Initial guess for the coordinates
-    x0 = np.random.random(len(matrix) * 2)
+    if x0 is None:
+        x0 = np.random.random(len(matrix) * 2)
```
sgrava:
```python
# TODO: allow to pass seed (see https://github.com/pasqal-io/qoolqit/issues/189)
rng = np.random.default_rng(0)

if x0 is None:
        # random initial positions
        x0 = rng.random((len(matrix), 2), dtype=matrix.dtype)

# make sure positions are of same type as matrix and flatten
x0 = np.asarray(x0.flatten(), dtype=matrix.dtype)
```

qoolqit/embedding/algorithms/interaction_embedding.py
```diff
@@ -19,9 +19,12 @@ class InteractionEmbedderConfig(EmbedderConfig):
     method: str = "Nelder-Mead"
     maxiter: int = 200000
     tol: float = 1e-8
+    x0: np.ndarray | None = None
 
 
-def interaction_embedding(matrix: np.ndarray, method: str, maxiter: int, tol: float) -> DataGraph:
+def interaction_embedding(
+    matrix: np.ndarray, method: str, maxiter: int, tol: float, x0: np.ndarray | None
```
sgrava:
```suggestion
    matrix: np.ndarray, method: str, x0: np.ndarray | None, maxiter: int, tol: float
```

qoolqit/embedding/algorithms/interaction_embedding.py
sgrava, line 38:
```suggestion
        method: the method used by scipy.minimize.
        x0: starting positions.
```

sgrava (changes requested):

sgrava:
Regarding naming, every embedder I saw has its own :) 
Options to normalize embedding starting positions:
- starting_positions
- pos (networkx use this)
- positions (blade use this)
- init_pos
- init_positions
- x0 (scipy.minimize)
- init ([smacof](https://scikit-learn.org/stable/modules/generated/sklearn.manifold.smacof.html) use this)

In the context of embedders I'm more in favour of being explicit, so variation on positions.
Let me know if you can think about other ones that could make sense. 
Let's pick one for every embedder.
We can postpone decision in another issue and keep x0 if you wish.

Edit: actually yes let's solve naming in another issue.

sgrava (approved):
Nice!

PR #267, opened 2026-04-15
Job api

### LINKED ISSUE #266: Job API

abussy-pasqal:
TODO:

- [ ] More tests
- [ ] Share with stakeholders
- [ ] Update documentation
- [ ] Answer questions in the comments
- [ ] Update notebooks
- [ ] Make a function to recreate a remote job from it's `id` and a connection (see `execution.ipynb`)


qoolqit/execution/job.py
```diff
@@ -0,0 +1,183 @@
...
+        """Return the current status of the job."""
+        ...
+
+    @abstractmethod
+    def id(self) -> str:
+        """Return the unique identifier of the job."""
+        ...
+
+    @abstractmethod
+    def result(self, timeout: float = math.inf, wait: float = 5) -> ResultType:
```
sgrava:
```suggestion
    def results(self, timeout: float = math.inf, wait: float = 5) -> ResultType:
```

qoolqit/execution/job.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def __init__(self, result: Results | None, message: str = "") -> None:
+        super().__init__()
+        self._result = result
+        self._message = message
+
+    def get_status(self) -> JobStatus:
+        return JobStatus.FAILED if self._result is None else JobStatus.SUCCEEDED
+
+    def id(self) -> str:
```
sgrava:
```suggestion
    def job_id(self) -> str:
```
just to make a clear distinction between `batch_id` and `job_id`

qoolqit/execution/job.py
```diff
@@ -0,0 +1,192 @@
...
+    """The job is currently being executed."""
+
+    SUCCEEDED = auto()
+    """The job completed successfully."""
+
+    FAILED = auto()
+    """The job terminated due to a failure."""
+
+    CANCELLED = auto()
+    """The job was cancelled before or during execution."""
```
sgrava: Would it make sense to just import `JobStatus` this from `pulser.backend.remote`?

sgrava (commented):
few first comments

qoolqit/execution/job.py
```diff
@@ -0,0 +1,192 @@
...
+    def __init__(self, remote_results: remote.RemoteResults) -> None:
+        super().__init__()
+        self._remote_results = remote_results
+        job_ids = self._remote_results.job_ids
+        assert len(job_ids) > 0
+        # If in open-batch mode, the job should be the last one in the batch ?
+        self._id: str = job_ids[-1]
+
+    def get_status(self) -> JobStatus:
+        status = self._remote_results.get_status()
```
sgrava: `_remote_results.get_status()` this one does not exist.
`_remote_results.get_batch_status()` will return the status of the batch.

To get the status of the job seems a bit more cumbersome.
Seems there is this
`_remote_results._connection._query_job_progress(
                self.batch_id
            ).items()`

qoolqit/execution/job.py
abussy-pasqal: It could. I'm matching the naming conventions from the proposal on Notion.

I have 2 problems with it though : 

1. `pulser.backend.remote.JobStatus.CANCELED` is misspelled
2. `DONE` has a different semantic here than in `pulser.backend.remote.JobStatus`. It means "terminal state" in the former, and "successful" in the latter.

sgrava: > 1. `pulser.backend.remote.JobStatus.CANCELED` is misspelled

It is just UK vs US English, so I'm fine with it.

> 2. `DONE` has a different semantic here than in `pulser.backend.remote.JobStatus`. It means "terminal state" in the former, and "successful" in the latter.

I see your point, so feel free to choose. I'm fine with both.
Personally, I would still prefer reusing Pulser code to avoid value maps and different meaning across the stack.

qoolqit/execution/backends.py
```diff
@@ -109,12 +108,16 @@ def __init__(
...
-        results = self._backend.run()
-        res_seq = (results,) if isinstance(results, Results) else tuple(results)
-        return res_seq
+        # Maybe don't catch the exception ?
+        try:
+            results = self._backend.run()
+            assert isinstance(results, Results)
+            return job._LocalJob(results, "")
+        except Exception as e:
+            return job._LocalJob(None, str(e))
```
sgrava:
```suggestion
        results = self._backend.run()
        assert isinstance(results, Results)
        return job._LocalJob(results, "")
```
I agree that catching a generic exception is not really useful. What do you think?

abussy-pasqal: The idea was to match an async behavior where the exceptions might not be easy to gather. But it's definitely premature to start storing the exception.

qoolqit/execution/job.py
a-quelle-pasqal: I'm inclined to agree with Stefano that reusing the existing enum would be preferable, as long as there is a value for each state we expect to have, which I'd expect to be the case.

qoolqit/execution/job.py
```diff
@@ -0,0 +1,192 @@
...
+
+class Job(ABC, Generic[ResultType]):
+    """Abstract representation of an asynchronous job or task.
+
+    A Job tracks the lifecycle of a unit of work submitted for execution.
+    It provides methods to inspect its state, wait for its result, and cancel it.
+
+    Lifecycle:
+        PENDING -> RUNNING -> SUCCEEDED
+                           -> FAILED
```
a-quelle-pasqal: If you do decide to use the job status from pulser, this docstring would need to be updated.

qoolqit/execution/backends.py
```diff
@@ -109,12 +108,12 @@ def __init__(
         self._backend_type = backend_type
         self._emulation_config = self.validate_emulation_config(emulation_config)
 
-    def run(self, program: QuantumProgram) -> Sequence[Results]:
+    def run(self, program: QuantumProgram) -> job.Job[Results]:
         """Run a compiled QuantumProgram and return the results."""
```
sgrava, line 112:
```suggestion
        """Run a compiled QuantumProgram and return a job handler.
        
        The returned handler `Job` can be used to retrieve results with `job.results()`
        
        Args:
            program (QuantumProgram): the compiled quantum program to run.
        """
```

sgrava (commented):
just a small one

abussy-pasqal:
Important points :

1. I opted for directly using the connection instead of `RemoteResults`. It felt more natural to me
2. I let some responsibilities to the connection : backoff, caching, ...
3. When converting from `RemoteResults`, I assumed that the job of interest would be the last in the list. Does it apply to open-batch ?
4. Batch related functionalities are implemented as free functions

Open questions :

1. As @sgrava and @a-quelle-pasqal mentioned, it might be a good idea to re-use the status naming from pulser and the cloud
2. `message()` is just a placeholder. Should we keep it ?

qoolqit/execution/backends.py
```diff
@@ -170,31 +169,23 @@ def __init__(
...
-        return remote_results
-
-    def run(self, program: QuantumProgram) -> Sequence[Results]:
-        """Run a compiled QuantumProgram remotely and return the results."""
-        remote_results = self.submit(program, wait=True)
-        res_seq: Sequence[Results] = remote_results.results
-        return res_seq
+        remote_results = self._backend.run(wait=False)
+        # If in open-batch mode, the job should be the last one in the batch ?
+        return job._RemoteJob._from_remote_results(remote_results, -1)
```
HGSilveri, line 188: Naive question because I'm not that familiar with the qoolqit codebase: can `RemoteResults` in this method ever have more than a single `Results` instance within them?

qoolqit/execution/job.py
```diff
@@ -0,0 +1,472 @@
...
+
+import math
+import time
+from abc import ABC, abstractmethod
+from enum import Enum, auto
+from typing import Generic, TypeVar
+
+from pulser.backend import Results, remote
+
+ResultType = TypeVar("ResultType")
```
HGSilveri, line 10: nit: I guess you can bind this to `Results`, right?
```suggestion
ResultType = TypeVar("ResultType", bound=Results)
```

qoolqit/execution/backends.py
```diff
@@ -109,12 +108,12 @@ def __init__(
...
 
-    def run(self, program: QuantumProgram) -> Sequence[Results]:
+    def run(self, program: QuantumProgram) -> job.Job[Results]:
         """Run a compiled QuantumProgram and return the results."""
         self._backend = self._backend_type(program.compiled_sequence, config=self._emulation_config)
         results = self._backend.run()
-        res_seq = (results,) if isinstance(results, Results) else tuple(results)
-        return res_seq
+        assert isinstance(results, Results)
+        return job._LocalJob(results)
```
HGSilveri, line 116: I'm a bit confused here. I thought one of the motivations behind the Job API was to have non-blocking calls to `run()`, regardless the backend, but this seems blocking to me. Am I missing or forgetting something from past discussions?

qoolqit/execution/job.py
abussy-pasqal: The concept of Job is independent of the result type. This is not (yet) relevent in our case. Maybe I can bind it here, and if we ever need a more general result type, we can factorize in with a `BaseJob` parent class ?

qoolqit/execution/backends.py
abussy-pasqal: I think there are restrictions on running multiple emulators (sv and mps) concurrently on the same system. Can you confirm @a-quelle-pasqal ? To be conservative, I implemented local jobs as blocking for now. This is implentation-defined though, and the API will already be there when we'll want to make a robust async implementation.

The other benefit is to have a common API between local and remote.

qoolqit/execution/backends.py
abussy-pasqal: qoolqit submits one job at a time to the wrapped backends. With `open_batch`, I believe that `RemoteResults` handle all the job of the batch. The assumption here is that the last submitted one will be at the end of the list. 

There might also be a race condition here if multiple jobs are submitted concurrently in the same batch.

Here is an example with `open_batch`

```python
from qoolqit.execution import QPU

def example():
    qpu = QPU(connection)
    with qpu._backend.open_batch() as opened_batch:
        while condition:
            job = qpu.run(program)
            # Query internals just for demonstration
            assert job._batch_id == opened_batch.backend._batch_id
            results = job.results()
            ...
```

HGSilveri: At the Pulser level, each call to `RemoteBackend.run()` returns a new `RemoteResults` instance. In the open batch case, every `RemoteResults` instance will have the same `batch_id` but different `job_ids`. 
So what you are doing seems fine to me

qoolqit/execution/job.py
HGSilveri: In that case it's best not to bind it at all and leave it as like this

qoolqit/execution/backends.py
abussy-pasqal: Are you certain ? If job ids aren't specified at the creation of `RemoteResults`, aren't all jobs fetched from the Cloud ?

HGSilveri: I'm certain that `PasqalCloud` [always specifies the subset of new job IDs](https://github.com/pasqal-io/pasqal-cloud/blob/fdd2f557fea12e69ac0c13d894595cc963851691/pulser-pasqal/pulser_pasqal/pasqal_cloud.py#L146) but it's true that doing so is not strictly required by the interface so other `RemoteConnection` subclasses may fail to do so.

sgrava (dismissed):
All good to me, nice `Job`!

HGSilveri:
> 2\. `message()` is just a placeholder. Should we keep it ?

Unless we have a use for it right now, I would remove it. It can always be added later if needed

sgrava (approved):
Nice!

Issue #279, opened 2026-04-21
support dmm


sgrava:
Support for Detuning Map Modulator (DMM) available in Pulser
 - https://pulser.readthedocs.io/en/stable/tutorials/dmm.html

## Description

Refine current (undocumented) solution, which lets the user define a `WeightedDetuning` to be passed to the drive. At compilation, this object is then converted into a DMM channel in Pulser.

### Actions
- [ ] deprecate `list[WeightedDetuning]` in drive
- [ ] Introduce `DetuningMapModulator` class, which take a strictly negative waveform and weights (same len as # of atoms in the register)
- [ ] document

PR #284, opened 2026-04-23
remove qoolqit utils

### LINKED ISSUE #283: remove qoolqit.utils
`qoolqit.utils` only contain some given tolerances and the `EQUAL` method.
 which can be replaced by numpy equivalent methods.

sgrava:
Resolves #283 

warm up cleaning before 
- refactoring drive
- refactoring WeightedDetunings
- add dmm support in #279 

## Changes
- removing qoolqit.utils and `EQUAL` method
- replacing it with equivalent `math.isclose()`


v-vitale (approved):
LGTM

PR #289, opened 2026-04-27
fix compilation error on interpolated waveform

### LINKED ISSUE #288: compilation error on a valid program
Programs with `Interpolated` waveform fails to compile on `DigitalAnalogDevice`.

### 1. The error in QoolQit
The following program fails to compile, but it is a valid one:
```python
from qoolqit import QuantumProgram, Register, Drive, Interpolated, DigitalAnalogDevice

register = Register.from_coordinates([[0,0],[0,1]])
amp_wave = Interpolated(60.0, [0.0, 0.5, 1.0, 0.5, 0.0])
drive = Drive(amplitude=amp_wave)
program = QuantumProgram(register=register, drive=drive)
program.compile_to(device=DigitalAnalogDevice())
```
and raises an unexpected pulser error:
```sh
ValueError: The pulse's amplitude goes over the maximum value allowed for the chosen channel.
```
### 2. The error in pulser
The equivalent pulser sequence, also fail for the same reason. To reproduce:
```python
from pulser import ConstantWaveform, Sequence, Register, DigitalAnalogDevice, InterpolatedWaveform, Pulse

max_amp = DigitalAnalogDevice.channels["rydberg_global"].max_amp

register = Register.from_coordinates([[0.0, 0.0], [0.0, 4.0]], prefix="q")
amp_wave = InterpolatedWaveform(60.0, [0.0, 0.5, max_amp, 0.5, 0.0])
det_wave = ConstantWaveform(60.0, 0.0)
pulse = Pulse(amplitude=amp_wave, detuning=det_wave, phase=0.0)
sequence = Sequence(register, DigitalAnalogDevice)
sequence.declare_channel("rydberg_global", "rydberg_global")
sequence.add(pulse,"rydberg_global")
```
raises:
```sh
File ~/Workspace/qoolqit/.venv/lib/python3.13/site-packages/pulser/channels/base_channel.py:484, in Channel.validate_pulse(self, pulse)
    482 amp_samples_np = pulse.amplitude.samples.as_array(detach=True)
    483 if self.max_amp is not None and np.any(amp_samples_np > self.max_amp):
--> [484](https://file+.vscode-resource.vscode-cdn.net/home/stefano/Workspace/qoolqit/dataset/~/Workspace/qoolqit/.venv/lib/python3.13/site-packages/pulser/channels/base_channel.py:484)     raise ValueError(
    485         "The pulse's amplitude goes over the maximum "
    486         "value allowed for the chosen channel."
    487     )
    488 if self.max_abs_detuning is not None and np.any(
    489     np.round(
    490         np.abs(pulse.detuning.samples.as_array(detach=True)),
   (...)    493     > self.max_abs_detuning
    494 ):
    495     raise ValueError(
    496         "The pulse's detuning values go out of the range "
    497         "allowed for the chosen channel."
    498     )

ValueError: The pulse's amplitude goes over the maximum value allowed for the chosen channel.
```
### 3. Minimal unexpected behaviour
Sampling the `InterpolatedWaveform` causes the round off error:

```python
from pulser import InterpolatedWaveform
max_amp = 15.707963267948966
max_wf_sampled = InterpolatedWaveform(60.0, [0.0, 0.5, max_amp, 0.5, 0.0]).samples.as_array().max()
print(max_wf_sampled)
print(max_wf_sampled > max_amp)
```
which outputs
```sh
15.707963268
True
```

## To solve
- [ ] temporarily patch 
- [x] raise an issue in pulser to understand if the behaviour is expected: https://github.com/pasqal-io/Pulser/issues/1051

sgrava:
Resolves #288 

## Changes
- unit test
- temporary fix: for interpolated waveforms, when translating the Interpolated waveform to the pulser equivalent, preemptively truncate values.

### DIFF
--- a/qoolqit/execution/compilation_functions.py
+++ b/qoolqit/execution/compilation_functions.py
@@ -60,10 +60,11 @@ class WaveformConverter:
         """Convert a QoolQit waveform into a equivalent Pulser waveform."""
         pulser_duration = self._pulser_duration(waveform)
 
+        # see https://github.com/pasqal-io/qoolqit/issues/288
+        # pulser.InterpolatedWaveform round values to 1e8 which can lead to amplitudes
+        # higher than what the device allows.
+        # Once solved in pulser the additional rounding below can be removed.
         if isinstance(waveform, Interpolated):
-            # fix https://github.com/pasqal-io/qoolqit/issues/288
-            # pulser.InterpolatedWaveform round values to 1e8 which
-            # can lead to higher amplitudes that what the device allows.
             self._energy = math.floor(self._energy * 1e8) / 1e8
 
         return waveform._to_pulser(duration=pulser_duration) * self._energy
--- a/qoolqit/execution/compilation_functions.py
+++ b/qoolqit/execution/compilation_functions.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 from enum import Enum
 
 from pulser.devices import Device as PulserDevice
@@ -13,6 +14,7 @@ from qoolqit.devices import Device
 from qoolqit.drive import Drive, Waveform, WeightedDetuning
 from qoolqit.exceptions import CompilationError
 from qoolqit.register import Register
+from qoolqit.waveforms import Interpolated
 
 
 class CompilerProfile(Enum):
@@ -57,6 +59,13 @@ class WaveformConverter:
     def convert(self, waveform: Waveform) -> ParamObj | PulserWaveform:
         """Convert a QoolQit waveform into a equivalent Pulser waveform."""
         pulser_duration = self._pulser_duration(waveform)
+
+        if isinstance(waveform, Interpolated):
+            # fix https://github.com/pasqal-io/qoolqit/issues/288
+            # pulser.InterpolatedWaveform round values to 1e8 which
+            # can lead to higher amplitudes that what the device allows.
+            self._energy = math.floor(self._energy * 1e8) / 1e8
+
         return waveform._to_pulser(duration=pulser_duration) * self._energy
 
 
--- a/qoolqit/execution/compilation_functions.py
+++ b/qoolqit/execution/compilation_functions.py
@@ -64,6 +64,7 @@ class WaveformConverter:
         # pulser.InterpolatedWaveform round values to 1e8 which can lead to amplitudes
         # higher than what the device allows.
         # Once solved in pulser the additional rounding below can be removed.
+        # see https://github.com/pasqal-io/Pulser/issues/1051
         if isinstance(waveform, Interpolated):
             self._energy = math.floor(self._energy * 1e8) / 1e8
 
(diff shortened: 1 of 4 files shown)


qoolqit/execution/compilation_functions.py
```diff
@@ -57,6 +59,14 @@ def _pulser_duration(self, waveform: Waveform) -> int:
     def convert(self, waveform: Waveform) -> ParamObj | PulserWaveform:
         """Convert a QoolQit waveform into a equivalent Pulser waveform."""
         pulser_duration = self._pulser_duration(waveform)
+
+        # see https://github.com/pasqal-io/qoolqit/issues/288
+        # pulser.InterpolatedWaveform round values to 1e8 which can lead to amplitudes
+        # higher than what the device allows.
+        # Once solved in pulser the additional rounding below can be removed.
+        if isinstance(waveform, Interpolated):
+            self._energy = math.floor(self._energy * 1e8) / 1e8
```
abussy-pasqal: I'm not an expert on numerical analysis. Aren't there corner cases for your formula ? I would be more comfortable with 
```python
max(energy - 1e-8, 0)
```

sgrava: Not an expert either but corner cases are for `self._energy`
- negative
- very very large
- special like np.inf

which are not an issue here because `self._energy` can only be positive. Moreover, if very, very large, compilation/sequence creation will fail anyway because the value of the amplitude is just to big.
Here I'm just trying to replicate the same rounding in pulser as a temporary fix [(see here)](https://github.com/pasqal-io/Pulser/blob/3ae06415543abfe99141990f108f54b45b64cb15/pulser-core/pulser/waveforms.py#L974),
which uses [numpy.round()](https://github.com/numpy/numpy/blob/v2.2.0/numpy/_core/fromnumeric.py#L3619)

abussy-pasqal: To handle negative numbers, you can use `math.trunc`. I agree with all the cases you mentioned, they are not relevent.

qoolqit/waveforms/waveforms.py
```diff
@@ -234,10 +234,19 @@ def min(self) -> float:
...
+    def _to_pulser(
+        self,
+        duration: int,
+        energy_factor: float = 1.0,
+    ) -> ParamObj | pulser.InterpolatedWaveform:
+        # Truncate the values to the specified energy factor,
+        # to avoid overflow when converting to pulser waveforms.
+        # see: https://github.com/pasqal-io/qoolqit/issues/288
+        # see: https://github.com/pasqal-io/Pulser/issues/1051
+        truncated_values = np.floor(self._values * energy_factor * 1e8) / 1e8
```
abussy-pasqal: Can `InterpolatedWaveform` be used with negative values ? e.g. detuning ?  If so, shouldn't you use `math.trunc` ? And add a test with negative values

abussy-pasqal (approved):

PR #295, opened 2026-04-30
add dmm

### LINKED ISSUE #294: add single dmm support

sgrava:
Resolves #294 
- [x] rename  `list[WeightedDetuning]` to `DetuningMapModulator`
- [x] check weights keys in {qubit_ids}, in program creation

## Changes
- rename `WeightedDetuning` -> `DetuningMapModulator`
- support only single `dmm: DetuningMapModulator`  instance in Drive
    context: real devices will only have one dmm channel. For emulation of multiple channels we will provide directly local addressability.
- tests

## Example
```python
from qoolqit import Register, Drive, QuantumProgram, DataGraph
from qoolqit.waveforms import Constant, PiecewiseLinear, Ramp, Interpolated
from qoolqit.drive import DetuningMapModulator
from qoolqit.devices import DigitalAnalogDevice

# Create the register
graph = DataGraph.circle(5)
register = Register.from_graph(embedded_graph)

# Defining the annealing parameters
omega = 0.8
delta_i = -2.0 * omega
delta_f = -delta_i
T = 25.0

amplitude = Interpolated(T, values=[0.0, omega/2, omega, omega/2, 0])
detuning = Ramp(T, delta_i, delta_f)
dmm = DetuningMapModulator(waveform=Constant(T,-1.0), weights={0:0.1})
drive = Drive(amplitude = amplitude, detuning=detuning, dmm=dmm)  
program = QuantumProgram(register=register, drive=drive)

# Compile the program to the device
program.compile_to(DigitalAnalogDevice())

# Draw the program
program.compiled_sequence.draw(draw_detuning_maps=True)
```
image
image

## Notes
- documentation and example in notebook to be added separately in #279

### DIFF
--- a/qoolqit/drive.py
+++ b/qoolqit/drive.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
+import warnings
 from dataclasses import dataclass
-from logging import warning
 from typing import Any, Sequence
 
 import matplotlib.pyplot as plt
@@ -131,7 +131,7 @@ class Drive:
 
         self._weighted_detunings = weighted_detunings
         if self._weighted_detunings is not None:
-            warning(
+            warnings.warn(
                 "Warning: `weighted_detunings` in Drive is deprecated. "
                 "Use `dmm` instead. This will be removed in a future version."
             )
@@ -182,10 +182,10 @@ class Drive:
         else:
             raise NotImplementedError(f"Composing with object of type {type(other)} not supported.")
 
-    def __amp_header__(self) -> str:  # pragma: no cover
+    def __amp_header__(self) -> str:
         return "Amplitude: \n"
 
-    def __det_header__(self) -> str:  # pragma: no cover
+    def __det_header__(self) -> str:
         return "Detuning: \n"
 
     def __repr__(self) -> str:
--- a/qoolqit/drive.py
+++ b/qoolqit/drive.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-import warnings
 from dataclasses import dataclass
-from typing import Any, Sequence
+from typing import Any
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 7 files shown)


abussy-pasqal:
Why deprecating `list[WeightedDetuning]` and instead of removing it ?

sgrava:
> Why deprecating `list[WeightedDetuning]` and instead of removing it ?

I kept it for the discussion, but I'm quite confident we can remove it. Let's sync about it.

v-vitale (dismissed):
It reads very well imho. I would remove the weighted_detunings from the Drive right away.

qoolqit/drive.py
```diff
@@ -41,7 +43,7 @@ def __init__(
         *,
         amplitude: Waveform,
         detuning: Waveform | None = None,
-        weighted_detunings: list[WeightedDetuning] | None = None,
+        dmm: DetuningMapModulator | None = None,
```
abussy-pasqal, line 46: I'm not a fan of `None` here. `DetuningMapModulator` already has a "no DMM" semantic, which is `weights = {}`. This would also require initializing waveform to the zero function by default.

sgrava: > DetuningMapModulator already has a "no DMM" semantic, which is weights = {}. This would also require initializing waveform to the zero function.

Entertaining this idea will cause other issues down the path I believe
It would require:
```
def __init__(
    self,
     ...
    dmm: DetuningMapModulator = DetuningMapModulator(Delay(duration=?, weights={}))
```
- decide a duration for the empty DMM. Which one?
- if the device does not support DMM, this will trigger an error. So we'll need to teach how to ignore the empty DMM.
- if the device supports it, it performs unnecessary manipulation.

abussy-pasqal: About points 2 and 3, you can filter at qoolqit's level, e.g. with something like : 
```python
class DetuningMapModulator:
  def enabled(self) -> bool:
      return len(self.weights) > 0
```
For point 1, I see your point. Duration doesn't really matter (0?) since dmm would be disabled though.

But tbh, this is a code philosophy and your PR is perfectly fine.

abussy-pasqal (approved):

PR #310, opened 2026-05-11
add  device type check and test

### LINKED ISSUE #309: type check device in compile_to
- type check `device` in `compile_to`, i.e. `isinstance(device, qoolqit.Device)`
- raise an error if device is not `qoolqit.Device`

Needed since it is currently easy to make the mistake and feed to the compilation a `pulser.Device` instead. 
Doing so, the raised error is not clear enough.

sgrava:
Resolves #309 

> - type check `device` in `compile_to`, i.e. `isinstance(device, qoolqit.Device)`
> - raise an error if device is not `qoolqit.Device`
>Needed since it is currently easy to make the mistake and feed to the compilation a `pulser.Device` instead. 
>Doing so, the raised error is not clear enough.

## Changes
- check 
- test


tests/test_compilation/test_compile_to.py
```diff
@@ -67,3 +68,14 @@ def test_compilation_with_dmm(profile: CompilerProfile) -> None:
...
+
+def test_compile_to_wrong_device_type() -> None:
+    """Test compilation to a device with a different type."""
+    register = Register(qubits={"q0": (0.0, 0.0)})
+    drive = Drive(amplitude=Constant(2.0, 0.2))
+    program = QuantumProgram(register=register, drive=drive)
+    with pytest.raises(TypeError, match="`device` must be of type `qoolqit.devices.Device`."):
+        program.compile_to(device="I'm not a device")  # type: ignore [arg-type]
+    with pytest.raises(TypeError, match="`device` must be of type `qoolqit.devices.Device`."):
+        program.compile_to(device=PulserAnalogDevice)  # type: ignore [arg-type]
```
abussy-pasqal, line 81:
```suggestion
        program.compile_to(device=PulserAnalogDevice())  # type: ignore [arg-type]
```

abussy-pasqal (changes requested):

tests/test_compilation/test_compile_to.py
sgrava: But pulser devices are not callable, right?
This test is to check that passing a pulser device instead of a `qoolqit` one is not allowed.
Maybe a small comment would help.

abussy-pasqal: My bad. I thought that `PulserAnalogDevice` was a `class`, same as qoolqit's

abussy-pasqal (approved):

PR #326, opened 2026-05-26
Custom weights graph indocumentation

v-vitale:
As per title, we add the example of a graph with custom weights on the edges.


sgrava (approved):
Nice!

PR #340, opened 2026-06-11
rename waveforms

### LINKED ISSUE #339: rename waveforms
Rename waveform classes to follow `pulser` naming conventions

## Description

The current waveform class names (`Constant`, `Ramp`, `Delay`, `Blackman`, `Sin`, etc.) are too short
and generic. They risk name collisions in user code, are harder to discover by name, and diverge from
the naming conventions used by the underlying `pulser` library, which these classes directly wrap.

## Current State

In `qoolqit/waveforms/waveforms.py`, the concrete waveform classes are currently named:

| Current name      | Maps to pulser type                        |
|-------------------|--------------------------------------------|
| `Constant`        | `pulser.ConstantWaveform`                  |
| `Ramp`            | `pulser.RampWaveform`                      |
| `Delay`           | `pulser.ConstantWaveform(duration, 0.0)`   |
| `Blackman`        | `pulser.BlackmanWaveform`                  |
| `Sin`             | *(custom)*                                 |
| `PiecewiseLinear` | *(composite of `Ramp`)*                    |
| `Interpolated`    | `pulser.InterpolatedWaveform`              |

## Proposed Change

Rename all waveform classes to include the `Waveform` suffix, mirroring the `pulser` naming convention:

| Current name      | Proposed name              |
|-------------------|----------------------------|
| `Constant`        | `ConstantWaveform`         |
| `Ramp`            | `RampWaveform`             |
| `Delay`           | `DelayWaveform`            |
| `Blackman`        | `BlackmanWaveform`         |
| `Sin`             | `SinWaveform`              |
| `PiecewiseLinear` | `PiecewiseLinearWaveform`  |
| `Interpolated`    | `InterpolatedWaveform`     |

## Motivation

- **Consistency with `pulser`**: `pulser` already uses `ConstantWaveform`, `RampWaveform`,
  `BlackmanWaveform`, `InterpolatedWaveform`. Mirroring this reduces cognitive overhead when switching
  between the two libraries.
- **Clarity**: A name like `ConstantWaveform` is unambiguous in user code — `Constant` could refer
  to anything.
- **Discoverability**: Users typing `*Waveform` or browsing autocomplete will find all waveform types
  grouped together.
- **Avoids name collisions**: Short names like `Ramp`, `Constant`, and `Sin` are common in scientific
  Python code and prone to shadowing.

## Affected Files

The rename touches multiple files across the codebase:

- `qoolqit/waveforms/waveforms.py` — class definitions
- `qoolqit/waveforms/__init__.py` — public exports
- `qoolqit/__init__.py` — top-level exports (`Blackman, Constant, Delay, Interpolated, PiecewiseLinear, Ramp, Sin`)
- `qoolqit/drive.py` — imports and docstring examples
- `tests/test_waveforms.py`, `tests/test_drive.py`, `tests/test_program.py`, `tests/conftest.py` — test imports
- Jupyter notebooks under `docs/` — tutorial code

## Backward Compatibility

To avoid a hard breaking change, the old names should be kept as deprecated aliases in
`qoolqit/waveforms/__init__.py` for one release cycle:

```python
# Deprecated aliases — will be removed in a future release
Constant = ConstantWaveform
Ramp = RampWaveform
Delay = DelayWaveform
Blackman = BlackmanWaveform
Sin = SinWaveform
PiecewiseLinear = PiecewiseLinearWaveform
Interpolated = InterpolatedWaveform
```

sgrava:
Resolves #339 

## Changes
- change waveform naming scheme
- deprecation mechanism

For example, the following import will trigger the warning
```python
>>> from qoolqit import Constant
.../qoolqit/qoolqit/__init__.py:79: DeprecationWarning: Constant is deprecated and will be removed in a future release. Use the equivalent ConstantWaveform instead.
.../qoolqit/qoolqit/__init__.py:79: DeprecationWarning: Constant is deprecated and will be removed in a future release. Use the equivalent ConstantWaveform instead.
```


tests/test_waveforms.py
```diff
@@ -8,29 +8,29 @@
 from numpy.typing import ArrayLike
 from scipy.integrate import quad
 
-from qoolqit import Blackman
```
abussy-pasqal, line 11: Could you add a test that checks that `from qoolqit import Constant` still works and emits a deprecation warning? I realize adding a dedicated test for this might be a bit overkill, but it would make the behavior explicit and protect against regressions.

```python
import sys
import pytest

def test_deprecated_import_constant_alias():
    # Fresh interpreter state for that module (helps avoid order-dependent failures)
    sys.modules.pop("qoolqit", None)

    ns = {}
    with pytest.warns(DeprecationWarning, match=r"Constant is deprecated"):
        exec("from qoolqit import Constant", ns)

    # Check the alias points to the replacement
    from qoolqit.waveforms import ConstantWaveform
    assert ns["Constant"] is ConstantWaveform
```

abussy-pasqal (dismissed):

tests/test_waveforms.py
sgrava: Good point! 
Added the test. It is only testing the `getattr` mechanism so I think is better scoped.

qoolqit/__init__.py
```diff
@@ -48,6 +48,9 @@
     "DigitalAnalogDevice",
     "MockDevice",
     "Device",
+    # Deprecated short-form aliases (e.g. "Blackman", "Constant", ...) are intentionally
+    # excluded from __all__. They are handled dynamically via __getattr__ below so that
+    # they do not appear in auto-complete or wildcard imports, discouraging new usage.
```
abussy-pasqal, line 53: Nice comment, I didn't know that !

abussy-pasqal (approved):

abussy-pasqal:
Note that this PR breaks:
```python
from qoolqit.waveforms import Constant
```

PR #343, opened 2026-06-15
remove sin and custom waveforms

### LINKED ISSUE #351: remove custom waveforms from docs
see #350

sgrava:
Part of #350 to refine the waveforms module
Resolves #351

## Description
Currently a program must be compiled to a `pulser.Sequence`, which implies that qoolqit waveforms must eventually be mapped to the waveforms available in pulser.
They should also unequivocally support rescaling (*), min and max methods because compilation heavily relies on those.

Custom made wavefoms like `Sin` are eventually mapped to an Interpolated waveform, but:
- number of samples to use is arbitrary
- no clear strategy to rescale, min and max method for arbirary functions, which are simply defined as callables.

It is simpler to just let the user define its own Interpolated waveform in the first place, with a clean 1 to 1 mapping to pulser.

For these reasons, we remove docs about making a custom waveform.

## Changes
- remove custom `Sin` waveform
- remove instructions on how to make a custom waveform inheriting from base `Waveform`

## Small fixes
- only define `__rshift__`

In `base_waveforms.py`, lines 120–127, both `__rshift__` / `__rrshift__` are defined.
```
def __rshift__(self, other: Waveform) -> CompositeWaveform:
    return self.__rrshift__(other)

def __rrshift__(self, other: Waveform) -> CompositeWaveform:
    if isinstance(other, Waveform):
        if isinstance(other, CompositeWaveform):
            return CompositeWaveform(self, *other._waveforms)  # ← self comes first
        return CompositeWaveform(self, other)
```
`__rshift__ (a >> b)` delegates to `__rrshift__`, which is semantically wrong. `__rrshift__` is Python's reflected right-shift, which calls `b.__rrshift__(a)`, i.e. the right (other) calls the method on left.
Internally this work because the `__rrshift__` overload swap the intended order.
 

## Future development
If this feature will be needed explicitly in the future and not covered by `Interpolated` case we can plan a solution.


v-vitale (approved):
LGTM!

Issue #350, opened 2026-06-16
refine wavefoms module


sgrava:
Collection of issues to refine the `qoolqit.waveforms` module for v1.2.
All waveforms must support rescaling, min, max and _to_pulser. This will allow to simplify compilation which heavily rely on these methods.

- [x] Interpolated only allows shape preserving interpolation
       #317 
       #320 
- [x] rename waveforms to match pulser scheme (`Constant` -> `ConstantWavefom`)
       #339 
       #340
- [x] remove Sin and custom waveforms from docs. User should just use `InterpolatedWaveform` or `PiecewiseLinearWaveform` to make custom waveforms
      #351
      #343 
- [x] simplify endpoint inclusion in composite waveforms
    #361
    #363 
- [x] set abstract methods min, max, _to_pulser in base wavefom. 
    #354 
    #355 
- [x] set abstract methods `__mul__` in base wavefom.
    #366 
    #367

PR #355, opened 2026-06-17
set min max _to_pulser abstract methods in base wavefom

### LINKED ISSUE #354: all waveforms must support rescaling min max and pulser conversion
Related to macro issue #350

1. All waveforms must support
    - `__mul__`, i.e. `ConstantWaveform(10.0,1.0)*3 = ConstantWaveform(10.0,3.0)`
    - `min`
    - `max`
    - `_to_pulser`

2. remove approximate min/max strategies as they can cause over/underestimating the minimum and maximum of a waveform.
QoolQit typically attempt to fully exploit device capability by rescaling waveforms to the device maximum allowed amplitude. The `max_amplitude` is a hard constrain on the device and underestimating the rescaling factor might cause compilation errors.

sgrava:
Following macro issue #350

Related to #354 

## Changes
- requires all waveforms to implement, `min`, `max`, `_to_pulser`
- remove approximate min/max methods, we cannot rely on such approximations becasue they can over/undershoot  values, while we work instead with stric boundaries. See #350.
- move and expand tests to `test_base_waveforms.py`

## Next
- `__mul__` will be handled in a separate PR, thus this does not close #354
Such feature is already supported but needs to be explicit in the base Waveform class


abussy-pasqal (approved):
Related, but not necessarily in the scope of this PR : 

Tests `test_blackman` and `test_piecewise` don't check that any value is within `min()` and `max()` (strictly, not up to a precision), unlike all the other waveforms tests in `test_waveforms.py`. I have https://github.com/pasqal-io/Pulser/issues/1051 in mind

sgrava:
> Related, but not necessarily in the scope of this PR :
> 
> Tests `test_blackman` and `test_piecewise` don't check that any value is within `min()` and `max()` (strictly, not up to a precision), unlike all the other waveforms tests in `test_waveforms.py`. I have [pasqal-io/Pulser#1051](https://github.com/pasqal-io/Pulser/issues/1051) in mind

Good point, I opened and issue about it #365

PR #358, opened 2026-06-19
rename single call and simplify logic

### LINKED ISSUE #357: simplify waveform single call
In `qoolqit/waveforms/`
- rename `__single_call__` to `_single_call` since it is an internal method rather than some python default method
- simplify logic

sgrava:
Sub issue from #350 
Resolves #357 

> In `qoolqit/waveforms/`
> - rename `__single_call__` to `_single_call` since it is an internal method rather than some python default method
> - simplify logic

### DIFF
--- a/qoolqit/waveforms/base_waveforms.py
+++ b/qoolqit/waveforms/base_waveforms.py
@@ -100,7 +100,7 @@ class Waveform(ABC):
             self._approximate_min_max()
         return cast(float, self._min)
 
-    def __single_call__(self, t: float) -> float:
+    def _single_call(self, t: float) -> float:
         return 0.0 if (t < 0.0 or t > self.duration) else self.function(t)
 
     @overload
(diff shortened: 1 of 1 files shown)


qoolqit/waveforms/base_waveforms.py
```diff
@@ -110,19 +110,11 @@ def __call__(self, t: float) -> float: ...
...
-            else:
-                raise TypeError(
-                    "Waveform array calling is supported on Python lists or NumPy arrays."
-                )
-            return value_array
-        else:
-            return self.__single_call__(t)
+        if isinstance(t, np.ndarray):
+            return np.vectorize(self._single_call)(t)
+        if isinstance(t, list):
```
abussy-pasqal, line 115: Have you considered converting the list to a `np.ndarray` ?

abussy-pasqal (approved):

qoolqit/waveforms/base_waveforms.py
sgrava: Yes it is a good point.
Let's rethink about this better when I'll introduce also torch parameters for waveforms.

PR #363, opened 2026-06-29
simplify composite waveform function

### LINKED ISSUE #361: simplify endpoint inclusion in composite waveforms
Remove dead code in `CompositeWaveform.function`

## Description
The logic inside `CompositeWaveform.function` can be simplified.
Only the last time of the last waveform needs to be treated separately since `np.seachsorted(side="right")` excludes it.
After that point the last waveform naturally evaluates to 0 outside bounds because of `Waveform.function` and `_single_call` implementation.

sgrava:
Resolves #361 

> Remove dead code in `CompositeWaveform.function`
> 
> ## Description
> The logic inside `CompositeWaveform.function` can be simplified.
> Only the last time of the last waveform needs to be treated separately since `np.seachsorted(side="right")` excludes it.
> After that point the last waveform naturally evaluates to 0 outside bounds because of `Waveform.function` and `_single_call` implementation.

No additional tests since it is just removing dead code

### DIFF
--- a/qoolqit/waveforms/base_waveforms.py
+++ b/qoolqit/waveforms/base_waveforms.py
@@ -237,8 +237,8 @@ class CompositeWaveform(Waveform):
             idx = idx - 1
 
         local_t = t - self.times[idx]
-        value = self.waveforms[idx](local_t)
-        return cast(float, value)
+        value: float = self.waveforms[idx](local_t)
+        return value
 
     def max(self) -> float:
         """Get the maximum value of the waveform."""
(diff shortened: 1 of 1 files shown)


qoolqit/waveforms/base_waveforms.py
```diff
@@ -230,15 +230,11 @@ def n_waveforms(self) -> int:
...
-        if idx == -1:
-            return 0.0
+
+        # clamp idx to the last waveform
         if idx == self.n_waveforms:
-            if t == self.times[-1]:
-                idx = idx - 1
-            else:
-                return 0.0
+            idx = idx - 1
```
abussy-pasqal: why not use `numpy.clip` or 
```python
idx = min(max(0, idx), self.n_waveforms - 1) 
```
?

sgrava: Nice, it looks cleaner with `np.clip`
Thanks!

abussy-pasqal (approved):

PR #367, opened 2026-06-30
add Waveform.__mul__ method

### LINKED ISSUE #365: test wavefoms samples within min max bounds
## Description
For example, tests `test_blackman` and `test_piecewise` don't check that any value is within `min()` and `max()` (strictly, not up to a precision), unlike all the other waveforms tests in `test_waveforms.py`.

## Acceptance criteria
- all waveforms should strictly check that sampled values are always within the waveform [min,max] bounds

### LINKED ISSUE #366: add rescaling abstract method in base waveform
Related to macro issue #350 

Base waveforms should support rescaling.

Add:
- [ ] abstract `__mul__` method in `Waveform`
- [ ] check that all waveforms support multiplication
- [ ] test all

sgrava:
Resolves #366 and #365

> 
> Related to macro issue #350 
> 
> Base waveforms should support rescaling.
> 
> Add:
> - [x] abstract `__mul__` method in `Waveform`
> - [x] check that all waveforms support multiplication
> - [x] test all

## Changes
- add `__mul__`/* abstract method
- add `__mul__`/* operator  to derived waveforms
- reshape and add `__mul__` tests
- adds tests to `BlackmanWaveform` and `PiecewiseWavaform` to test that their sampling respect min/max bounds of the waveform

### DIFF
--- a/qoolqit/waveforms/waveforms.py
+++ b/qoolqit/waveforms/waveforms.py
@@ -25,6 +25,9 @@ class DelayWaveform(Waveform):
     def min(self) -> float:
         return 0.0
 
+    def __mul__(self, other: float) -> Waveform:
+        return self
+
     def _to_pulser(self, duration: int) -> ParamObj | pulser.ConstantWaveform:
         return pulser.ConstantWaveform(duration, 0.0)
 
@@ -59,6 +62,13 @@ class RampWaveform(Waveform):
     def min(self) -> float:
         return min([self.initial_value, self.final_value])
 
+    def __mul__(self, other: float) -> Waveform:
+        return RampWaveform(
+            self.duration,
+            initial_value=self.initial_value * other,
+            final_value=self.final_value * other,
+        )
+
     def _to_pulser(self, duration: int) -> ParamObj | pulser.RampWaveform:
         return pulser.RampWaveform(duration, self.initial_value, self.final_value)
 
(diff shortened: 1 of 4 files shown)


qoolqit/waveforms/waveforms.py
```diff
@@ -253,6 +271,9 @@ def min(self) -> float:
     def max(self) -> float:
         return float(self._values.max())
 
+    def __mul__(self, other: float) -> Waveform:
+        return InterpolatedWaveform(self.duration, self._values * other, self._times)
```
abussy-pasqal, line 275: does `PchipInterpolator` guarantee homogeneity ? In other words, is the scaling exact for all `t`, or only at `self._times` and an approximation at other points ?

Could you please add a comment on this ?

sgrava: Hi! good question.
Not mentioned explicitly but I'd say that the sacaling is the same for the whole duration, just by construction. 
Its piecewise cubic intepolants are entirely fixed by times/values.

Anyway a test is better than an argument so I'll add something to check it.

sgrava: Added a very minimal check!

```
    # check that also samples are scaled correctly
    times = np.linspace(0, wf.duration, 17)
    wf_samples = wf(times)
    scaled_samples = scaled(times)
    np.testing.assert_allclose(scaled_samples, wf_samples * scaling, atol=1e-8)
```

sgrava: Regarding the comment I think it belong more to the abstract `__mul__`. The sense being that it is really the contract of multiplication to homogeneously rescale all waveform.

abussy-pasqal (approved):

Issue #250, opened 2026-04-09
convert all markdown-exec to notebooks


sgrava:
See #226 

> `mardown-exec` is used extensively in the documentation but its future as a plugin for both `mkdocs` or `properdocs` is not clear.
> To relax constrains on the decision about the documentation build backend:
> - [ ] convert all `mardown-exec` instances in docs to notebooks and do not depend any more on `mardown-exec`

PR #372, opened 2026-07-01
remove docs utils

### LINKED ISSUE #371: remove unused docs/utils.py

sgrava:
Resolves #371 

file `utils.py` contains a method to convert figures to HTML.
This is not used anymore since we moved away from the `markdown-exec` extension (see #250 ).


v-vitale (approved):
LGTM

PR #373, opened 2026-07-02
refine docs qoolqit model page 1/2

### LINKED ISSUE #370: refine docs/get_started/qoolqit_model 1/2
Review `docs/get_started/qoolqit_model` to fix typos and inconsistencies.

- [x]  add paragraph structure at the beginning; title can be “general model”
- [x]  $t$ → $\tilde{t}$ in $\tilde{H}$
- [ ]  ~~add explanation of occupation operator~~
- [x]  add definition of $r_{ij}$ adimensional distance between atom i and atom j sites before $J_{i,j}$
- [x]  just use the table and get rid of verbose description that can go into description column
    - [x]  for $\tilde{J}$, put in the table the reference to r^-6, and leave to a new line after the table the fact that the user must use values smaller than or equal to 1.
        - [x]  SHALL we explain that the value max =1 is because it corresponds to the the smallest physical distance that can be realized in the machine?
        - [x]  NB there is also an inconsistency : first we say the max $J_{ij}= 1$ then we say $J_{ij} \leq 1$
- [ ]  ~~Create a table listing the meaningful max parameters: (e.g. $\tilde J_{\text{max}} = \max_{i<j}\tilde J_{ij} \leq 1$ and $\tilde \Omega_{max} = \max_{\tilde{t}}\tilde\Omega$~~
    - [x]  note we have to decide the value of $\tilde J_{\text{max}}  \;\leq\; 1$ or $=1$
    - [ ]  ~~if we show device spec with $\tilde{J}=1$ or equivalently at the scale $r=r_\text{min}^\text{dev}$, then we should be consistent probably with this~~
    - [ ]  ~~a possible solution is to mention here $\max \tilde{J}=1$ and leave for an advanced section the possibility of having also $\leq 1$?~~
    - [ ]  ~~Possible notation abuse: meaning of $\tilde{J}$ and the energy scale. The former is about the values entered by the user and we are trying to use it also to talk about the energy scale, meaning what is 1 in energy scale.~~

sgrava:
Resolves #370 

- resolves small actionables issues in #370 
- reshape `qoolqit_model.md` into two main topics: qoolqit dimensionless model and compilation


docs/get_started/qoolqit_model.md
```diff
@@ -1,24 +1,22 @@
...
+- Drive strength and time regimes in the dynamics
+- The meaning of compilation
 
-As a result, the behavior of the system is not set by absolute values alone, but by the **interplay between geometry (distances) and control strength (laser power)**. Different combinations of these quantities can lead to equivalent physical behavior, as long as their relative scales are preserved.
-
-!!! info "Take-home message 1"
-    QoolQit introduces a **dimensionless reference frame** where all quantities are expressed in function of an **interaction reference**.
-
-The quantum system is thus described by the sum of two energetic contributions, the interaction and the driving one, as described by the following Hamiltonian:
+In Rydberg neutral-atom systems, atoms interact through a combination of **distance-dependent interactions** and **laser-driven controls**, as described by the following dimensionless Hamiltonian:
```
ferrulli1pasqal, line 11: At this moment in the formula below it is not clear what J^d_max is. 
It is explained only after and it is also not very visually clear.

docs/get_started/qoolqit_model.md
```diff
@@ -31,37 +29,35 @@ $$
...
 | $\tilde{t}$ | Dimensionless time | $> 0$ |
 
-The introduced many-body Hamiltonian has rich dynamics, resulting from the interplay between the driving and interaction terms over time.
-To help users understand how to define a concrete program, we briefly describe below the expected physical regimes for particular choices of driving strength (amplitude) and program duration. We will see that their values relative to the program's maximum interaction strength,
+We call a **quantum program** the collection of all these parameters, which fully specify the time dependent Hamiltonian defined above.
 
-$$
-\tilde J_{\text{max}} \;=\; \max_{i<j}\tilde J_{ij} \;\leq\; 1,
-$$
+Importantly, in QoolQit, the dimensionless Hamiltonian is obtained dividing by $J_{\text{max}}^{d}$, the maximum available coupling strength that can be realized on a hardware device.
```
ferrulli1pasqal: it sounds hard the coupling strength. Maybe just few words to clarify better.
Also, we should mention more clearly that this value is realised using the r^d_min (minimum radius of the device).

docs/get_started/qoolqit_model.md
```diff
@@ -1,24 +1,22 @@
...
 \left(
-\cos\phi(t)\,\hat{\sigma}^x_i - \sin\phi(t)\,\hat{\sigma}^y_i
+\cos(\phi) \hat{\sigma}^x_i - \sin(\phi)\hat{\sigma}^y_i
 \right)}_{\text{global drive}}
 -
-\underbrace{\sum_i \left( \tilde{\delta}(t) + \epsilon_i\,\tilde{\Delta}(t) \right) \hat{n}_i}_{\text{detuning}}.
+\underbrace{\sum_i \left( \tilde{\delta}(\tilde{t}) + \epsilon_i\,\tilde{\Delta}(\tilde{t}) \right) \hat{n}_i}_{\text{detuning}}.
 $$
 
 Here, $\hat{n}_i = \frac{1}{2}(1 + \hat{\sigma}^z_i)$ is the Rydberg occupation operator of atom $i$, and the $\hat{\sigma}^{x,y,z}_i$ are the Pauli operators:
```
ferrulli1pasqal: can we include the notation for n_i in terms of braket? I personally find it much more intuitive, but is personal taste

docs/get_started/qoolqit_model.md
```diff
@@ -31,37 +29,35 @@ $$
...
+Importantly, in QoolQit, the dimensionless Hamiltonian is obtained dividing by $J_{\text{max}}^{d}$, the maximum available coupling strength that can be realized on a hardware device.
 
-is what matters.
+!!! info "Take-home message 1"
+    QoolQit introduces a **dimensionless model** where all quantities are expressed relative to an **interaction reference**.
+
+Such reference makes the program definition hardware independent and has several advantages:
+- Remove hardware constants: By defining a new unit of energy all device-dependent constants are factored out.
+- Dimensionless parameters: dimensionless drive and interaction strengths, allows to easily explore different physical regimes.
+- Hardware-agnostic algorithm development: developers can build algorithms/programs focusing on ideas rather then hardware specifications.
```
ferrulli1pasqal: add portability as well. The algorithm uses the relative HW information, then the real physical quantities are computed evrey time a specific device is specified.

docs/get_started/qoolqit_model.md
```diff
@@ -31,37 +29,35 @@ $$
...
+
+Moreover, on the practical side, since $\tilde{J}_{ij}=1/\tilde{r}_{ij}^{6}$ follows Rydberg scaling and can be at most equal to $1$, also $\tilde{r}_{min}\geq 1$, i.e. the minimum pairwise distance that can be realized is always $1$.
+Finally, also time is made dimensionless, $\tilde{t}=tJ_{max}^{d}$, to realize an equivalent dynamics.
+
+To help users understand how to define a concrete program, we briefly describe below the expected physical regimes for particular choices of driving strength (amplitude) and program duration. We will see that their values relative to the program's maximum interaction strength, $J_{\text{max}} \leq 1$, is what matters.
```
ferrulli1pasqal: To me is not clear what J_max is? (also maybe tilde is missing?)

docs/get_started/qoolqit_model.md
```diff
@@ -89,39 +85,43 @@ Next we will discuss the compilation, the crucial step to translate a QoolQit di
...
     Compilation is the step where QoolQit takes a dimensionless program and rescales it into hardware-realizable parameters, while preserving physical invariants.
 
+Practically, as anticipated in the previous section, a device choice will set the interaction energy reference.
 
-A convenient way to understand this is to first work entirely in dimensionless units. As mentioned above, the key idea is that the program is defined by **ratios**, not by absolute scales. For example, fixing the ratio $\frac{\max_{\tilde{t}}\tilde{\Omega}}{\tilde{J}}$ defines a line in the $(\tilde{J},\tilde{\Omega})$ plane. Moving along this line changes the overall scale of the program, but preserves its dimensionless structure (here $\max_{\tilde{t}}$ stands for the maximum over time).
-
-This means that compilation does **not** change the dimensionless physics of the program. Instead, it rescales the program so that it lies inside the region that can be implemented on a given device.
+!!! info "Take-home message 2"
+    The actual physical scale $J_{\text{max}}^{d}$, such as the precise distances or laser amplitudes, is determined during the **compilation step**, when targeting a specific device.
```
ferrulli1pasqal: this is not true ?!?! This is the value fixed by using r_d_min (see comments above as well), non?

docs/get_started/qoolqit_model.md
```diff
@@ -89,39 +85,43 @@ Next we will discuss the compilation, the crucial step to translate a QoolQit di
...
 
+More details about the connection to physical units are provided in the section [Adimensionalization and Compilation](../extended_usage/adimensionalization.md).
 
 ### Example
```
ferrulli1pasqal: I would create a new page called compilation examples, or so. Or just delete this part as the same info are somewhere else and once we will define the new default as working point this example will not be anymore relevant here.

ferrulli1pasqal (commented):
first iteration

docs/get_started/qoolqit_model.md
sgrava: yep tilde missing, fixed it

docs/get_started/qoolqit_model.md
sgrava: added

docs/get_started/qoolqit_model.md
sgrava: Let's discuss it in another PR. I actually prefer this one instead.

docs/get_started/qoolqit_model.md
sgrava: Let's discuss this in a separate issue/PR 
#377

docs/get_started/qoolqit_model.md
sgrava: Everything is explained after presenting the Hamiltonian. Please, specify where do you prefer $J_{\text{max}}^d$ to appear

docs/get_started/qoolqit_model.md
sgrava: coupling strength -> interaction strenght

>Also, we should mention more clearly that this value is realised using the r^d_min (minimum radius of the device).

done

docs/get_started/qoolqit_model.md
sgrava: outdated, this text was removed in the new version

docs/fundamentals/compilation/rationale.md
```diff
@@ -83,6 +83,42 @@ In this regime the compiled register uses the smallest physical spacing the devi
...
 
+#### Example
+
+As mentioned above, compilation does **not** change the physics of the program. Instead, it rescales the program so that it lies inside the region that can be implemented on a given device.
+
+Consider the figure below:
+
+![Compilation diagram](../../extras/assets/compilation.svg)
+
+The valid parameters region (green box) of a device is constrained by $\tilde{J} \leq 1, \;\tilde{\Omega} \leq 0.2$. The bound $\tilde{J} \leq 1$ is compatible with a minimum spacing $a$ allowed in the register distance equal to $a_{\text{min}}=1$.
```
v-vitale: I don't know whether the bounds will be different in the future but the fact that $\tilde{J} \leq 1$ holds is by choice, while \;\tilde{\Omega} \leq 0.2$ might be different.

v-vitale (changes requested):

docs/get_started/qoolqit_model.md
```diff
@@ -83,54 +89,77 @@ In an interacting many-body system, time can be naturally measured relative to t
...
-
-However, the values that can actually be implemented are constrained by the **hardware**. Real devices only allow certain ranges of interaction strengths, drive amplitudes, detunings, and evolution times. Therefore, an important task of QoolQit is to take the dimensionless program specified by the user and map it to a set of parameters that can be realized on the chosen hardware. We refer to this step as **compilation**.
+A QoolQit program is expressed in dimensionless units, allowing users to define problems independently of any specific hardware platform.
+In contrast, hardware specifications are given in physical units and and can only realize a limited range of parameter values.
+Thus, before executing it on real hardware, a program must be compiled in two steps:
 
+- **Dimensionalization**: Conversion of dimensionless parameters into physical units of energy, distances and times, to fall within the capabilities of the target device (if possible).
+- **Translation**: Conversion into a lower-level representation that can directly instruct the target hardware (e.g., a [Pulser sequence](https://docs.pasqal.com/pulser/)).
```
ferrulli1pasqal: Let's just use **Conversion**: and remove Translation.

Ok splitting compilation in two phases. There is still left though the differentiation between adimensional -> dimensional and then the optimality based on rules.

ferrulli1pasqal (commented):
just a minor comment, otherwise it flows very well

docs/fundamentals/compilation/rationale.md
sgrava: Changed from

>The valid parameters region (green box) of a device is constrained by $\tilde{J} \leq 1, \;\tilde{\Omega} \leq 0.2$. The bound $\tilde{J} \leq 1$ is compatible with a minimum spacing $a$ allowed in the register distance equal to $a_{\text{min}}=1$.

to

The green box highlights the valid parameter region for the interaction energy $\tilde{J}\_{ij}$ and the driving amplitude $\tilde{\Omega}$.
As described in the [QoolQit model](../../get_started/qoolqit_model.md) page, the interaction energy is bounded by 1 by construction: QoolQit's adimensionalization enforces $\tilde{J}\_{ij} = J\_{ij}/J_{max}^{d} \leq 1$ across all devices.
The driving amplitude (more precisely, its maximum over time) is instead constrained to a device-dependent upper bound, $\tilde{\Omega} = \Omega/J_{max}^{d} \lesssim \Omega_{max}^{d}/J_{max}^{d} = 0.2$.

A bit more verbose but the entire fundamentals/compilation will get a restyle in a separate PR so fine for now.

docs/get_started/qoolqit_model.md
sgrava: Translation/compilation is better in my opinion because it implies a semantic-preserving process. It is more precise than conversion.

sgrava: rephrased a bit as

> 
> - **Dimensionalization**: Converts dimensionless parameters into physical units of energy, distance, and time so that they are compatible with the capabilities of the target device, when possible.
> - **Translation**: Compiles the program into a lower-level representation that can directly instruct the target hardware (e.g., a [Pulser sequence](https://docs.pasqal.com/pulser/)).

v-vitale (approved):

ferrulli1pasqal (approved):

PR #384, opened 2026-07-08
refine docs fundamentals compilation

### LINKED ISSUE #381: refine docs/fundamentals/compilation
Part of docs review #369 
- [ ]  Remove all refernce to $J_0$, it should be $J_{max}^{d}$ when it specifically refer to the device max    
- [ ]  ~~I would move the working point compilation in the advanced section~~ Both compilation should be documented here such that when we will chage the code in the future it will be easier to restructure docs.
- [ ]  in hardware modulation a picture that shows visually (same as in linked section). Show modulated sequence.

sgrava:
Resolves #381 

> - [x]  Remove all reference to $J_0$, it should be $J_{max}^{d}$ when it specifically refer to the device max    
> - [x]  ~~I would move the working point compilation in the advanced section~~ Both compilation should be documented here such that when we will chage the code in the future it will be easier to restructure docs.
> - [ ]  in hardware modulation a picture that shows visually (same as in linked section). Show modulated sequence.

Refines `docs/compilation/rationale.md`
- simplify compilation ideas, making it more clear that dimensionalization and profiles are different steps
- small working point description (true to actual code, we will change it in the future)
- single image, single explanation of `max_energy` example

In `docs/extended_usage/adimensionalization.md`
- Removing time regime interpretation. It is already displayed in `docs/get_started/qoolqit_model.md`


docs/fundamentals/compilation/rationale.md
```diff
@@ -1,144 +1,96 @@
...
-
-### Working point compilation
-In progress...
 
+The dimensionless content is unchanged: the ratio between drive and interaction is the same, and therefore the underlying physics encoded in the program is the same.
 
+### Working point
+The working point does not apply any rescaling on top of the dimensionalization of the quantum program.
+Unlike the default maximum energy profile, it preserves the user-chosen physical scales instead of maximizing the device energy scale.
+It is designed for users who really want to control the precise physical values of drive amplitude, detuning, distances, and time, and opt out of the default compilation profile.
```
v-vitale, line 84: I would add the image of before when the user choices a point in the region or outside the region and the one outside the region is marked with an x to say that it could not be compiled.

What do you think, is it clear? Btw we can add it at a later stage if you prefer. In order to release v1.2 and proceed with more structural changes.

v-vitale (dismissed):

ferrulli1pasqal (approved):

docs/fundamentals/compilation/rationale.md
sgrava: Ah nice suggestion.
For the moment, I simply added a new sentence pointing to the figure above.

>In terms of the figure above, this is equivalent to fixing a point in the $(\tilde{J}, \tilde{\Omega})$ plane: if that point lies inside the green region, or more generally satisfies the device specifications, the program will compile.

Do you think it is enough for now?
We can always add the figure you suggest later.

v-vitale (approved):
LGTM

PR #391, opened 2026-07-14
add docs preview

sgrava:
Resolves #389 

> add `pr-preview-action` to pipeline
> 
> ## Description
> Currently, the [docs stage](https://github.com/pasqal-io/qoolqit/blob/main/.github/workflows/docs.yml) only tests that the documentation succesfully builds.
> In the same workflow, we can also add the [pr-preview-action](https://github.com/rossjrw/pr-preview-action).
> This would:
> - test that the docs succesfully build
> - preview the built docs in github PRs

Issue #389, opened 2026-07-13
add pr-preview-action to pipeline


sgrava:
add `pr-preview-action` to pipeline

## Description
Currently, the [docs stage](https://github.com/pasqal-io/qoolqit/blob/main/.github/workflows/docs.yml) only tests that the documentation succesfully builds.
In the same workflow, we can also add the [pr-preview-action](https://github.com/rossjrw/pr-preview-action).
This would:
- test that the docs succesfully build
- preview the built docs in github PRs

PR #396, opened 2026-07-16
nbstripout and execute notebooks in mkdocs

### LINKED ISSUE #395: clean notebooks and run them when building the documentation
- add `nb-clean` to pre-commit job in the linting github workflow
- set `execute=true` in `mkdocs-jupyter` config in `mkodcs.yml`

## Description
In the documentation we heavily rely on notebooks, they are nice to automatically generate plots.
However, notebooks are difficult to review because they generate huge non-readable git diff.

To solve
1. strip all notebooks output and some metadata while pushing to the repository
2. run them when building the documentation

Preview is now generated automatically on every PR (see #389  #391 ) so it is inspectable.
Pipeline job will still fail if a notobook does not execute without errors.

sgrava:
Resolves #395

> - add `nb-clean` to pre-commit job in the linting github workflow
> - set `execute=true` in `mkdocs-jupyter` config in `mkodcs.yml`
> 
> ## Description
> In the documentation we heavily rely on notebooks, they are nice to automatically generate plots.
> However, notebooks are difficult to review because they generate huge non-readable git diff.
> 
> To solve
> 1. strip all notebooks output and some metadata while pushing to the repository
> 2. run them when building the documentation
> 
> Preview is now generated automatically on every PR (see #389  #391 ) so it is inspectable.
> Pipeline job will still fail if a notobook does not execute without errors.


v-vitale (approved):
Nice!

Issue #388, opened 2026-07-13
refine register module


sgrava:
Collection of register module issues:
- [x]  ~~Deprecate `Register.from_graph()`. Getting the coordinates or the `Register` directly from the `DataGraph` is cleaner.~~ Not really needed as long as we offer and present aternatives
- [x]  Graph → register labeling
       Asses how labels are handled? Is it coherent between graph and register? QoolQit register has to follow QoolQit labeling. It will be QoolQit responsibility to support conversion to Pulser labeling
    - `networkx` default label nodes as 0,1,2,3,4….. But in principle you can put whatever you want.
    - Pulser label atoms/nodes as “q0”, “q1”, “q2”, etc…

    Internally, each `networkx` node i becomes atom “q_i”, so order is preserved.
    
- [x]  Convenient methods for quick register creations.
       Same methods as `DataGraph`, which will change to only have edge weights, the register method ones will actually create coordinates.
    #403 

- [x] Center register drawing

- [x] `Register.__init__` lacks validation
     #397 
     #398
- [x] add `interaction_matrix` method
    #417 
    #419

sgrava:
Closing this leaving #403 open to be closed later.

PR #398, opened 2026-07-16
validate input in register default constructor

### LINKED ISSUE #397: add input validation to register default constructor
`Register.__init__` lacks input validation

**File:** `qoolqit/register.py` — `Register.__init__`

## Problems

1. **No empty dict guard**: `Register({})` is accepted silently, but `min_distance()`, `max_radial_distance()`, and `draw()` all crash downstream with unhelpful errors.

2. **No coordinate value validation**: values are never checked to be 2-tuples of numbers. Inputs like `{"a": "bad"}` or `{"a": (1, 2, 3)}` pass construction and cause cryptic errors later.

3. **Missing generic type annotation**: `self._qubits: dict` gives no type information to static checkers. Should be `dict[Any, tuple[float, float]]` or a named alias.

sgrava:
Resolves #397 

> `Register.__init__` lacks input validation
> 
> **File:** `qoolqit/register.py` — `Register.__init__`
> 
> ## Problems
> 
> 1. **No empty dict guard**: `Register({})` is accepted silently, but `min_distance()`, `max_radial_distance()`, and `draw()` all crash downstream with unhelpful errors.
> 
> 2. **No coordinate value validation**: values are never checked to be 2-tuples of numbers. Inputs like `{"a": "bad"}` or `{"a": (1, 2, 3)}` pass construction and cause cryptic errors later.
> 
> 3. **Missing generic type annotation**: `self._qubits: dict` gives no type information to static checkers. Should be `dict[Any, tuple[float, float]]` or a named alias.
> 

## Changes
- expand register docstrings
- validate input, we accept all cordinates that can be casted to numpy or torch array of shape (2,)

## Next
Next steps are collected in #388 
- validate input and small fixes in Register.from_coordinates()
- deprecate Register.from_graph()

### DIFF
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -7,6 +7,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib.figure import Figure
 from matplotlib.ticker import MultipleLocator
+from numpy.typing import ArrayLike
 
 from qoolqit.graphs import DataGraph, all_node_pairs, distances
 from qoolqit.graphs.utils import radial_distances
@@ -46,7 +47,9 @@ class Register:
         >>> reg = Register({"a": np.array([0.0, 0.0]), "b": np.array([1.0, 0.0])})
     """
 
-    def __init__(self, qubits: dict) -> None:
+    def __init__(
+        self, qubits: dict[str, ArrayLike | torch.Tensor] | dict[int, ArrayLike | torch.Tensor]
+    ) -> None:
         """Default constructor for the Register.
 
         Args:
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -22,20 +22,20 @@ except ImportError:
 
 
 class Register:
-    """The Register in QoolQit, representing a set of qubit ids with coordinates.
+    """A QoolQit register mapping qubit IDs to 2D coordinates.
 
     Attributes:
-        qubits: a dictionary of qubit ids and respective coordinates {q: (x, y),...}.
-        qubits_ids: a list of qubit ids.
+        qubits: a dictionary of qubit IDs and respective coordinates {q: (x, y),...}.
+        qubits_ids: a list of qubit IDs.
         n_qubits: the number of qubits in the register.
 
     Examples:
-        From a dictionary of qubit ids and coordinates:
+        From a dictionary of qubit IDs and coordinates:
 
         >>> reg = Register({"a": (0.0, 0.0), "b": (1.0, 0.0), "c": (0.0, 1.0)})
         >>> reg = Register({0: (0.0, 0.0), 1: (1.0, 0.0), 2: (0.0, 1.0)})
 
-        From a list of coordinates (qubit ids are assigned automatically as integers):
+        From a list of coordinates (qubit IDs are assigned automatically as integers):
 
         >>> reg = Register.from_coordinates([(0.0, 0.0), (1.0, 0.0), (0.0, 1.0)])
 
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from math import ceil
+from typing import TYPE_CHECKING
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -10,13 +11,23 @@ from matplotlib.ticker import MultipleLocator
 from qoolqit.graphs import DataGraph, all_node_pairs, distances
 from qoolqit.graphs.utils import radial_distances
 
+if TYPE_CHECKING:
+    import torch
+
+try:
+    import torch
+
+    _has_torch = True
+except ImportError:
+    _has_torch = False
+
 
 class Register:
     """The Register in QoolQit, representing a set of qubits with coordinates.
 
     Attributes:
         qubits: a dictionary of qubits and respective coordinates {q: (x, y),...}.
-        qubit_ids: a list of qubit ids.
+        qubits_ids: a list of qubit ids.
         n_qubits: the number of qubits in the register.
 
     Examples:
@@ -44,7 +55,8 @@ class Register:
         Raises:
             TypeError: If the qubits argument is not a dictionary.
             ValueError: If the qubits argument is empty.
-            ValueError: If the coordinate for a qubit is not a tuple, list or array of length 2.
+            ValueError: If the coordinate for a qubit is not a tuple, list or
+                numpy or torch array of length 2.
         """
         if not isinstance(qubits, dict):
             raise TypeError(
@@ -54,21 +66,26 @@ class Register:
         if len(qubits) == 0:
             raise ValueError("Register cannot be empty.")
 
+        parsed = {}
         for key, val in qubits.items():
-            val_array = val
-            if isinstance(val, (tuple, list)):
-                val_array = np.asarray(val, dtype=float)
-            valid = val_array.ndim == 1 and val_array.shape[0] == 2
-            if not valid:
+            try:
+                if _has_torch and isinstance(val, torch.Tensor):
+                    parsed[key] = torch.asarray(val, dtype=torch.float64)
+                else:
+                    parsed[key] = np.asarray(val, dtype=float)
+            except (ValueError, TypeError) as err:
                 raise ValueError(
-                    f"Coordinate for qubit {key!r} must be a tuple, list, or array of length 2, "
-                    f"got {val!r}."
+                    f"Coordinate for qubit {key!r} must be castable to an array of floats."
+                ) from err
+
+            converted = parsed[key]
+            if converted.ndim != 1 or converted.shape[0] != 2:
+                raise ValueError(
+                    f"Coordinate for qubit {key!r} must be a 1D "
+                    f"tuple, list, or array of length 2, got {val!r}."
                 )
 
-        self._qubits = {
-            key: (np.asarray(val, dtype=float) if isinstance(val, (tuple, list)) else val)
-            for key, val in qubits.items()
-        }
+        self._qubits = parsed
 
     @classmethod
     def from_graph(cls, graph: DataGraph) -> Register:
@@ -114,7 +131,7 @@ class Register:
     @property
     def n_qubits(self) -> int:
         """Number of qubits in the Register."""
-        return len(self.qubits)
+        return len(self._qubits)
 
     def distances(self) -> dict:
         """Distance between each qubit pair."""
@@ -151,7 +168,14 @@ class Register:
         fig, ax = plt.subplots(1, 1, figsize=(4, 4), dpi=150)
         ax.set_aspect("equal")
 
-        x_coords, y_coords = zip(*self.qubits.values())
+        x_coords, y_coords = zip(
+            *[
+                np.asarray(
+                    v.detach() if (_has_torch and isinstance(v, torch.Tensor)) else v, dtype=float
+                )
+                for v in self.qubits.values()
+            ]
+        )
         x_min, x_max = min(x_coords), max(x_coords)
         y_min, y_max = min(y_coords), max(y_coords)
 
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -51,7 +51,8 @@ class Register:
     """
 
     def __init__(
-        self, qubits: dict[str, ArrayLike | torch.Tensor] | dict[int, ArrayLike | torch.Tensor]
+        self,
+        qubits: dict[str, ArrayLike | torch.Tensor] | dict[int, ArrayLike | torch.Tensor],
     ) -> None:
         """Default constructor for the Register.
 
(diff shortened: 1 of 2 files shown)


qoolqit/register.py
```diff
@@ -1,28 +1,93 @@
...
         if not isinstance(qubits, dict):
-            raise TypeError(
-                "Register must be initialized with a dictionary of "
-                "qubits and respective coordinates {q: (x, y), ...}."
-            )
-
-        self._qubits: dict = qubits
+            raise TypeError("`qubits` must be a dictionary mapping qubit ids to coordinates.")
+        if len(qubits) == 0:
+            raise ValueError("Register cannot be empty.")
```
abussy-pasqal, line 100: I can't see why a register could not be empty. To me, it seems that a register is simply a decorated `dict`. So i would try to implement every function that is already implemented as its `dict` counterpart.

I'll add comments where relevent.

This raises the question of `Register` being only an alias of a `dict`, but this would be too huge a refactoring here

qoolqit/register.py
```diff
@@ -49,15 +114,16 @@ def from_coordinates(cls, coords: list) -> Register:
...
             )
         coords_dict = {i: pos for i, pos in enumerate(coords)}
         return cls(coords_dict)
 
     @property
     def qubits(self) -> dict:
-        """Returns the dictionary of qubits and respective coordinates."""
-        return self._qubits
+        """Returns a copy of the dictionary of qubits and respective coordinates."""
+        return dict(self._qubits)
```
abussy-pasqal: Please add a test that the 2D-vectors are indeed copied. Pseudo-code :

```python
qubits1 = register.qubits
qubits2 = register.qubits
qubits1["0"][0] = 1.0
assert qubits1 != qubits2
```

I suspect that this test will fail.

Avoid using `copy.deepcopy`. Use :
```python
import numpy as np
import torch
from copy import deepcopy

def copy_value(v):
    if isinstance(v, np.ndarray):
        return v.copy()
    if isinstance(v, torch.Tensor):
        # Don't use .detach() if you want to preserve autograd history,
        # which is probably not the case here
        return v.detach().clone()
    return deepcopy(v)

copied = {k: copy_value(v) for k, v in self._qubits.items()}
```

qoolqit/register.py
```diff
@@ -67,7 +133,7 @@ def qubits_ids(self) -> list:
     @property
     def n_qubits(self) -> int:
         """Number of qubits in the Register."""
-        return len(self.qubits)
+        return len(self._qubits)
```
abussy-pasqal: I would implement both `__bool__` and `__len__`:
```python
def __bool__(self) -> bool:
    return bool(self._qubits)

def __len__(self) -> int:
    return len(self._qubits)
```
This property would become
```python
def n_qubits(self) -> int:
    return len(self)
```

qoolqit/register.py
```diff
@@ -1,28 +1,93 @@
...
+        if len(qubits) == 0:
+            raise ValueError("Register cannot be empty.")
+
+        parsed = {}
+        for key, val in qubits.items():
+            try:
+                if _has_torch and isinstance(val, torch.Tensor):
+                    parsed[key] = torch.asarray(val, dtype=torch.float64)
+                else:
+                    parsed[key] = np.asarray(val, dtype=float)
```
abussy-pasqal: why keeping points as `torch` and not converting everything to `numpy` ? This would save a lot of double handling

abussy-pasqal: TBH, from a geometrical point of view, I would rather store a list of labels and a Nx2 `np.array` (vertices).
This would simplify a lot of the code :  most of (if not all) computations would become simple `numpy` functions

qoolqit/register.py
sgrava: > I can't see why a register could not be empty. To me, it seems that a register is simply a decorated dict. So i would try to implement every function that is already implemented as its dict counterpart.

I gave it a lot of thoughts about this.
Ultimately, my take is that an empty register is an invalid state (for what we want to achieve here), and shouldn't be propagated down the stack. 
I am happy to rethink about it if a concrete use case pops up.  What do you think?

qoolqit/register.py
sgrava: Nice suggestion! 
I added a `_copy` method and the the test you suggested.

qoolqit/register.py
sgrava: > why keeping points as torch and not converting everything to numpy ? This would save a lot of double handling

We want to keep the option to pass torch tensor as other users might come from there.
It was already allowed by the register constructor, just not validated, which is the point of this PR.
I'd like to keep it for now. Thanks to your suggestions is manageble and can be refactored easily.

> TBH, from a geometrical point of view, I would rather store a list of labels and a Nx2 np.array (vertices).
This would simplify a lot of the code :  most of (if not all) computations would become simple numpy functions

Good point. 
I store cordinates as an array now. 
- We can think later about touching the default constructor. A better solution would be to have `Register.from_dict` and `Register.from_coords`. Would you agree?
- we can think later about better leveraging this within the class methods. (This PR is already huge.)

qoolqit/register.py
sgrava: Done, have a look!

qoolqit/register.py
abussy-pasqal: It’s a perfectly valid choice.

That said, I think the two points you mentioned are worth distinguishing:

1. **“An empty register is an invalid state”**: this could be represented with a `__bool__` operator that returns `False` if and only if the register is empty.

2. **“It shouldn’t be propagated down the stack”**: one could argue that a register is a container, much like a `dict`. For some algorithms, an empty `dict` would be invalid input; for others, it would simply be a no-op. In my opinion, that decision should be left to the consumer.

So I agree that, in the absence of a concrete use case, enforcing the check in the constructor is probably the more convenient and safer option.

qoolqit/register.py
```diff
@@ -1,34 +1,150 @@
...
+        return x.clone()
+    return np.copy(x)
 
 
 class Register:
-    """The Register in QoolQit, representing a set of qubits with coordinates."""
+    """A QoolQit register mapping qubit IDs to 2D coordinates.
+
+    Attributes:
+        qubits: a dictionary of qubit IDs and respective coordinates {q: (x, y),...}.
```
abussy-pasqal: Shouldn't it be removed ?

qoolqit/register.py
```diff
@@ -1,34 +1,150 @@
...
+
+        >>> import torch
+        >>> reg = Register({"a": torch.tensor([0.0, 0.0]), "b": torch.tensor([1.0, 0.0])})
+    """
+
+    def __init__(
+        self,
+        qubits: (
+            Mapping[str, Sequence[float] | npt.NDArray[np.float64] | torch.Tensor]
+            | Mapping[int, Sequence[float] | npt.NDArray[np.float64] | torch.Tensor]
```
abussy-pasqal, line 82: why the duplication ?

qoolqit/register.py
```diff
@@ -1,34 +1,150 @@
...
-                "Register must be initialized with a dictionary of "
-                "qubits and respective coordinates {q: (x, y), ...}."
-            )
+        if not isinstance(qubits, Mapping):
+            raise TypeError("`qubits` must be a Mapping of qubit ids to coordinates.")
+        if not qubits:
+            raise ValueError("Register cannot be empty.")
+
+        self._qubits_ids: tuple[str | int, ...] = tuple(qubits.keys())
+        validated_coords = [self._validate_coord(k, c) for k, c in qubits.items()]
```
abussy-pasqal, line 103:
```suggestion
        validated_coords = [self._validate_coord(k, c).tolist() for k, c in qubits.items()]
```

abussy-pasqal: This would simplify the code of `_stack_coords`

qoolqit/register.py
```diff
@@ -109,13 +237,12 @@ def draw(self, ax: Axes | None = None, marker_size: int = 100) -> None:
         marker_radius = marker_size**0.5 / 2  # in points
         annotation_offset = 1.5 * marker_radius  # place label just outside the marker
 
-        x, y = zip(*self.qubits.values())
-        ax.scatter(x, y, s=marker_size, color="green")
-
-        for i, qubit_id in enumerate(self.qubits_ids):
+        coords = self._coords.detach().cpu().numpy() if _is_torch(self._coords) else self._coords
+        for xi, yi, qid in zip(coords[:, 0], coords[:, 1], self.qubits_ids):
+            ax.scatter(xi, yi, s=marker_size, color="green")
```
abussy-pasqal, line 239: Note that `scatter` accepts an array so you could do

```python
coords = self._coords.detach().cpu().numpy() if _is_torch(self._coords) else self._coords

ax.scatter(
    coords[:, 0],
    coords[:, 1],
    s=marker_size,
    color="green",
)
```
But `annotate` doesn't...

qoolqit/register.py
```diff
@@ -1,34 +1,150 @@
...
+def _norm(x: npt.NDArray[np.float64] | torch.Tensor) -> np.float64 | torch.Tensor:
+    if _is_torch(x):
+        norm: torch.Tensor = torch.linalg.norm(x)
+        return norm
+    return np.linalg.norm(x)
+
+
+def _copy(x: npt.NDArray | torch.Tensor) -> npt.NDArray | torch.Tensor:
+    if _is_torch(x):
+        return x.clone()
```
abussy-pasqal: add a a comment why you don't use `detach()` if that's intended

qoolqit/register.py
abussy-pasqal: Totally agree 👍

qoolqit/register.py
sgrava: You are right, removed them. Just delegate to property docstrings.

qoolqit/register.py
sgrava: Unfortunately, `annotate` is not vectorized.

qoolqit/register.py
sgrava: You are right, added a detach.

qoolqit/register.py
sgrava: I'm not a typing expert but to me those are just hints to the the user.
It is saying don't do weird combination of labels like [1, "q", 12, ...], just stick to a single type.

qoolqit/register.py
```diff
@@ -1,34 +1,145 @@
...
+    """
+    if _is_torch(val):
+        return val.to(dtype=torch.float64)
+    return np.asarray(val, dtype=np.float64)
+
+
+def _norm(x: npt.NDArray[np.float64] | torch.Tensor) -> np.float64 | torch.Tensor:
+    if _is_torch(x):
+        norm: torch.Tensor = torch.linalg.norm(x)
+        return norm
```
abussy-pasqal:
```suggestion
        return norm.item()
```

qoolqit/register.py
sgrava: You'll lose the numpy/torch information this way.
I think the best way to simplify here would be to make `_to_array`  bit smarter and recursively call itself to end up with an array idependently of the input.

qoolqit/register.py
abussy-pasqal: Sorry, I misread the nesting of `[]`. Maybe it's worth doing type aliases ?
```python
from typing import TypeAlias

Array: TypeAlias = npt.NDArray[np.float64] | torch.Tensor
ArrayLike: TypeAlias = npt.ArrayLike | Array
```
Since it's used a lot.

qoolqit/register.py
abussy-pasqal: With this, the function always returns a float (maybe you'll need casting)

sgrava:
@abussy-pasqal 

> Maybe it's worth doing type aliases ?
> 
> from typing import TypeAlias
> 
> Array: TypeAlias = npt.NDArray[np.float64] | torch.Tensor
> ArrayLike: TypeAlias = npt.ArrayLike | Array
> 
> Since it's used a lot.

Yes, let's explore it in #406

abussy-pasqal (dismissed):
Feel free to discard the last comment

qoolqit/register.py
sgrava: Done! Yes item is typed loosely. I cast to float.

qoolqit/register.py
```diff
@@ -39,11 +39,11 @@ def _to_array(
...
 
 
-def _norm(x: npt.NDArray[np.float64] | torch.Tensor) -> np.float64 | torch.Tensor:
+def _norm(x: npt.NDArray[np.float64] | torch.Tensor) -> float:
     if _is_torch(x):
-        norm: torch.Tensor = torch.linalg.norm(x)
-        return norm
-    return np.linalg.norm(x)
+        norm = torch.linalg.norm(x)
+    norm = np.linalg.norm(x)
```
abussy-pasqal: `np.linalg.norm` is also called in the `torch` branch

abussy-pasqal (changes requested):

abussy-pasqal (approved):

qoolqit/register.py
sgrava: ah my bad, thanks!

PR #405, opened 2026-07-27
Added Summation feature

### LINKED ISSUE #404: Add pointwise summation of waveforms
Waveforms can currently only be composed sequentially, with >> (CompositeWaveform), which concatenates them in time. 

There is no way to combine them simultaneously. This adds __add__, the pointwise sum of waveforms: (F + G)(t) = F(t) + G(t). Scalars are promoted to a constant waveform, so a constant offset on top of a shaped drive can be written directly:

`H= F + G`
`H = 0.15 + 3.0 * G`

`>>` keeps its meaning (concatenation in time); + adds values at each instant.

(only waveforms with the same duration can be added)

alessandro-santini:


v-vitale (changes requested):
Tests to be added, right?

PR #412, opened 2026-07-29
Added circle register classmethod

### LINKED ISSUE #407: add circle class method to Register
Add `Register.circle` class method to create a register which atoms are placed along the perimeter of a circle.

## Example
An easy implementation without `numpy` could be:

```python
@classmethod
def circle(
    cls,
    n: int,
    spacing: float = 1.0,
) -> Register:
    """Constructs a circle register.

    Args:
        n: number of qubits.
        spacing: distance between each qubit.
    """
    step = 2.0 * math.pi / n
    r = spacing / (2.0 * math.sin(math.pi / n))
    coords = [(math.cos(step * i) * r, math.sin(step * i) * r) for i in range(n)]

    return cls.from_coordinates(coords)
```

## Acceptance criteria
- spacing kwarg
- test simple layout with explicit coords
- test minimum distance is given by spacing

alessandro-santini:
Adds a Register.circle() classmethod that arranges n qubits evenly around a circle with a given spacing between adjacent qubits, computing the radius via the standard formula and validating inputs (n >= 1, spacing > 0).

Includes tests covering exact coordinates, input validation errors, and a min-distance check confirming the generated geometry actually matches the requested spacing.

### DIFF
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -179,6 +179,25 @@ class Register:
         coords_dict = {i: pos for i, pos in enumerate(coords)}
         return cls(coords_dict)
 
+    @classmethod
+    def circle(cls, n_qubits: int, spacing: float = 1.0) -> Register:
+        """Initializes a Register with qubits arranged in a circle.
+
+        Args:
+            n_qubits: number of qubits to place in the circle.
+            spacing: distance between adjacent qubits. Defaults to 1.0.
+        """
+        if n_qubits < 1:
+            raise ValueError("Number of qubits must be at least 1.")
+        if spacing <= 0:
+            raise ValueError("Spacing must be positive.")
+        if n_qubits == 1:
+            return cls.from_coordinates([(0.0, 0.0)])
+        radius = spacing / (2 * np.sin(np.pi / n_qubits))
+        angles = np.linspace(0, 2 * np.pi, n_qubits, endpoint=False)
+        coords = [(radius * np.cos(a), radius * np.sin(a)) for a in angles]
+        return cls.from_coordinates(coords)
+
     @property
     def qubits(self) -> dict:
         """Returns a dictionary of qubits and respective coordinates."""
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 from collections.abc import Mapping, Sequence
 from typing import TYPE_CHECKING, Any, TypeGuard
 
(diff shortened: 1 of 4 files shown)


qoolqit/register.py
```diff
@@ -220,13 +239,16 @@ def interactions(self) -> dict:
...
+    def draw(
+        self, ax: Axes | None = None, marker_size: int = 100, node_color: str = "green"
+    ) -> None:
         """Draw the register.
 
         Args:
             ax: an optional matplotlib Axes instance to draw on.
                 If None, a new Axes will be created.
             marker_size: size of the qubit markers in points squared. Defaults to 100.
+            node_color: color of the qubit markers. Defaults to "green".
```
sgrava: Nice! 
Just to keep it clean I suggest doing this separately in the other PR #416

tests/test_register.py
```diff
@@ -184,3 +184,20 @@ def test_draw() -> None:
...
+    radius = spacing / (2 * np.sin(np.pi / n_qubits))
+    expected = [
+        (radius, 0.0),
+        (0.0, radius),
+        (-radius, 0.0),
+        (0.0, -radius),
+    ]
+
+    assert register.n_qubits == n_qubits
+    np.testing.assert_allclose(list(register.qubits.values()), expected, atol=1e-8)
```
sgrava:
```suggestion
    np.testing.assert_allclose(register._coords, expected, atol=1e-8)
```
eventually this will become also a public method but you can get directly the (N,2) array of coordinates

tests/test_register.py
sgrava, line 1: Please, also check that the minimum distance is ~ spacing with a tolerance. Something like this

```python
@pytest.mark.parametrize("n, spacing", [(2, 1.0), (5, 1.28), ... other examples])
def test_circle_min_distance(n, spacing):
    register = Register.circle(n_qubits, spacing=spacing)
    assert np.testing.assert_allclose(register.min_distance, spacing, atol=1e-8)
```

sgrava, line 1: catch all invalid branches in separate test, i.e.
```python
def test_0_qubits():
   with pytest.raises(ValueError, match="Spacing must be positive."):
        Register.circle(2, spacing=-1)
```

qoolqit/register.py
```diff
@@ -179,6 +179,25 @@ def from_coordinates(
         coords_dict = {i: pos for i, pos in enumerate(coords)}
         return cls(coords_dict)
 
+    @classmethod
+    def circle(cls, n_qubits: int, spacing: float = 1.0) -> Register:
```
sgrava:
```suggestion
    def circle(cls, n: int, spacing: float = 1.0) -> Register:
```
`n_qubits` shadows a bit the property `register.n_qubits`. Maybe just n is fine, I guess it depends a bit on the argument naming conventions in the other methods.

qoolqit/register.py
```diff
@@ -179,6 +179,25 @@ def from_coordinates(
...
+        if n_qubits < 1:
+            raise ValueError("Number of qubits must be at least 1.")
+        if spacing <= 0:
+            raise ValueError("Spacing must be positive.")
+        if n_qubits == 1:
+            return cls.from_coordinates([(0.0, 0.0)])
+        radius = spacing / (2 * np.sin(np.pi / n_qubits))
+        angles = np.linspace(0, 2 * np.pi, n_qubits, endpoint=False)
+        coords = [(radius * np.cos(a), radius * np.sin(a)) for a in angles]
+        return cls.from_coordinates(coords)
```
sgrava, line 202: Still prefer the suggested implementation. 
There is no performance issues here a and it is more flexible, because the type of coords is determined by `spacing`.

sgrava (commented):
Looking good, a few comments.
Please, also fill out the PR description. Something minimal is fine.

sgrava (approved):
Nice!

PR #414, opened 2026-07-29
Add triangular classmethod to register

### LINKED ISSUE #409: add triangular class method to Register
Add `Register.triangular` class method to create a register which atoms are placed on a triangular lattice.

## Example
An easy implementation without `numpy` could be:

```python
@classmethod
def triangular(
    cls,
    rows: int,
    cols: int,
    spacing: float = 1.0,
) -> Register:
    """Constructs a triangular lattice register.

    Qubits are arranged in a 2D triangular lattice where odd rows are
    offset by half the column spacing, so that every qubit has
    equilateral-triangle nearest-neighbour distances equal to `spacing`.

    Args:
        rows: number of qubit rows.
        cols: number of qubit columns.
        spacing: distance between neighbouring qubits. Defaults to 1.0.

    Raises:
        ValueError: If `rows` or `cols` is less than 1.
    """
    if rows < 1 or cols < 1:
        raise ValueError("`rows` and `cols` must both be at least 1.")

    row_height = spacing * math.sqrt(3) / 2.0
    coords = [
        (col * spacing + (row % 2) * spacing / 2.0, row * row_height)
        for row in range(rows)
        for col in range(cols)
    ]
    return cls.from_coordinates(coords)
```

## Acceptance criteria
- spacing kwarg
- coordinates centered around the origin (0, 0)
- test simple layout with explicit coords
- test minimum distance is given by spacing

## Notes
The current implementation on `qoolqit/register.py` is not yet centered on the origin — needs an offset applied before this is closed.

alessandro-santini:
Adds a `Register.triangular()` classmethod that arranges qubits on a `rows × atoms_per_row` triangular lattice with a given spacing between adjacent qubits. Rows are staggered by half a spacing (odd rows shifted right), the layout is centred on the centroid of the qubits, and inputs are validated (`rows >= 1`, `atoms_per_row >= 1`, `spacing > 0`).
 
The signature and geometry match `pulser.Register.triangular_lattice()`.

Includes tests covering exact coordinates for both even and odd row counts, the row staggering and centring convention, input validation errors, and min-distance checks confirming the generated geometry actually matches the requested spacing.

### DIFF
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -189,6 +189,10 @@ class Register:
             n: number of columns of triangles.
             spacing: distance between adjacent qubits. Defaults to 1.0.
         """
+        if m < 1 or n < 1:
+            raise ValueError("`m` and `n` must both be at least 1.")
+        if spacing <= 0:
+            raise ValueError("`spacing` must be positive.")
         graph = nx.triangular_lattice_graph(m, n, with_positions=True)
         pos = nx.get_node_attributes(graph, "pos")
         coords = np.array([(x * spacing, y * spacing) for x, y in pos.values()])
--- a/qoolqit/register.py
+++ b/qoolqit/register.py
@@ -181,33 +181,27 @@ class Register:
         return cls(coords_dict)
 
     @classmethod
-    def triangular(cls, m: int, n: int, spacing: float = 1.0) -> Register:
-        """Constructs a triangular lattice register.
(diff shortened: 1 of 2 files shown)


v-vitale (dismissed):
Looks good.

I think we should also write a method for creating a triangular lattice by fixing the number of atoms in total instead of the number of cells, but I don't know what could be a smart way to differentiate between the two, i.e. how to call them.

sgrava (approved):
Nice

Issue #406, opened 2026-07-28
explore refactoring array handling in a separate module


sgrava:
Explore refactoring array handling in a separate module. 
- Ideally similar to [`pulser.math`](https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/math/__init__.py) 
- Type aliases could simplify type hints
   
```python
from typing import TypeAlias

Array: TypeAlias = npt.NDArray[np.float64] | torch.Tensor
ArrayLike: TypeAlias = npt.ArrayLike | Array
```
Benefits:
- reusing numpy/torch array handling in other parts of the library, mostly waveforms

PR #419, opened 2026-07-30
add interaction matrix method to register

### LINKED ISSUE #417: add interaction_matrix method to Register
add `interaction_matrix` method to Register

##  Description
From a given register, the `interaction_matrix` method should return a symmetric matrix encoding the Rydberg interactions $J_{ij}=1/r_{ij}^6$, with $r_{ij}=|r_{i}-r_{j}|$

sgrava:
Resolves #417 

> add `interaction_matrix` method to Register
> 
> ##  Description
> From a given register, the `interaction_matrix` method should return a symmetric matrix encoding the Rydberg interactions $J_{ij}=1/r_{ij}^6$, with $r_{ij}=|r_{i}-r_{j}|$

## Changes
- add `Register.interaction_matrix` method which return the Rydberg interaction matrix. Dispatches to `torch` or `numpy` depending on the coordinates type stored in the register
- test

## Bonus
- adds `matplotlib.use("Agg")` in `conftest.py`. Totally unrelated but let the tests to avoid loading matplotlib in interacting mode. Saves 0.5 seconds per test module.

## Next
- refactor abstract array interface #406


v-vitale (commented):
Just a comment.

What is the reason behind using torch tensors for the interaction matrix?

sgrava:
> Just a comment.
> 
> What is the reason behind using torch tensors for the interaction matrix?

It would be nice to support torch backed registers as well, but not a priority for now I agree

qoolqit/register.py
```diff
@@ -310,6 +317,22 @@ def interactions(self) -> dict:
...
+    def interaction_matrix(self) -> npt.NDArray[np.float64] | torch.Tensor:
+        """Interaction 1/r^6 between each qubit pair, as a matrix (0 on the diagonal)."""
+        dist_matrix = _pdist(self._coords)
+
+        # Avoid division-by-zero on the diagonal (where r_ii == 0):
+        # - torch: use a boolean diagonal mask and `torch.where` (out-of-place, autograd-safe).
+        # - numpy: set dist_matrix diagonal to 1.0 before exponentiation, then back to 0.0.
+        if _is_torch(dist_matrix):
+            diagonal_mask = torch.eye(dist_matrix.shape[0], dtype=torch.bool)
+            return torch.where(diagonal_mask, 0.0, dist_matrix ** (-6))
```
abussy-pasqal: no need to mask, dividing by zero with torch won't emit a warning

sgrava: Yes but it is more really to sanitize the output and return 0 on the diagonal by convention.
The `torch.fill_diagonal_` method is not really an alternative since it writes in-place.

Do you see an alternative here? Keeping the convention to have  0 on the diagonal I mean

abussy-pasqal: you can fill the diagonal with zeros after the computation

abussy-pasqal (dismissed):

qoolqit/register.py
sgrava: I think it is more clear now that filling the diagonal of a torch tensor needs to be out-of-place. Please, have another look!

abussy-pasqal (approved):

PR #430, opened 2026-08-05
Add to_matrix method to DataGraph

### LINKED ISSUE #429: add to_matrix method in DataGraph
Add a `to_matrix` method to `DataGraph` that returns the graph's connectivity as a real, symmetric, square matrix — the counterpart to the existing `DataGraph.from_matrix` classmethod (`qoolqit/graphs/data_graph.py`).

**Behavior**

- Returns an `(N, N)` `numpy.ndarray` where `N = number_of_nodes()`.
- Off-diagonal entry `(i, j)` / `(j, i)`:
  - `0` if no edge `(i, j)` exists.
  - The edge weight if edge `(i, j)` exists and has a weight set.
  - `1` if edge `(i, j)` exists but its weight is `None` (unweighted edge).
- Since `DataGraph` does not support self-loops, the diagonal is repurposed to store node weights: entry `(i, i)` is the weight of node `i`, or `0` if the node's weight is `None`.
- Result must be symmetric by construction.

**Round-trip**

`DataGraph.from_matrix(g.to_matrix())` should reproduce a graph equivalent to `g` (same nodes, edges, node/edge weights), matching the semantics documented on `from_matrix`.

**Notes**

- Node ordering in the matrix should follow `sorted(self.nodes)` for a deterministic, reproducible layout.
- Add tests covering: unweighted graphs, node-weighted graphs, edge-weighted graphs, and the `from_matrix`/`to_matrix` round trip.

sgrava:
## Summary
- Adds `DataGraph.to_matrix()`, the inverse of the existing `from_matrix` classmethod: returns a real symmetric square matrix with node weights on the diagonal (0.0 if unset, since self-loops aren't supported) and edge weights off-diagonal (defaulting to 1.0 for edges without an explicit weight).
- Nodes are mapped to matrix indices `0..N-1` following `self.nodes` insertion order (not label value), so it also works for non-integer or non-contiguous node labels.

## Known follow-up
- While testing, found that `DataGraph._edge_weights` isn't refreshed when edges are added after construction via `add_edges_from`/`add_edge` (only `_reset_dicts()` populates it, and that only runs inside the classmethod constructors). This doesn't affect `to_matrix` itself, but it's a pre-existing gap in the graph, tracked separately in #431.

## Tests
- [x] Unit tests for unweighted graphs, weighted graphs, custom/non-integer node labels with `None` weights, and a `to_matrix`/`from_matrix` round trip.

Closes #429

## Alternatives
`networkx.adjacency_matrix` is a good ready-made alternative but we cannot use it because of bug #431

### DIFF
--- a/qoolqit/graphs/data_graph.py
+++ b/qoolqit/graphs/data_graph.py
@@ -295,29 +295,26 @@ class DataGraph(BaseGraph):
         return graph
 
     def to_matrix(self) -> np.ndarray:
-        """Return the graph as a real symmetric square matrix.
+        """Return the connectivity matrix of this graph.
 
-        The inverse of `from_matrix`. Node weights are stored in the diagonal,
-        since self-loops are not supported. For each edge (i, j), the entries
-        M[i, j] and M[j, i] are set to its weight, or to 1.0 if the edge has
-        no weight set. Missing node weights are set to 0.0 in the diagonal.
-
-        Nodes are ordered by sorting `self.nodes`.
+        The inverse of `from_matrix`.
+        Nodes are mapped to indices 0,..N-1 according to `self.nodes` order.
+        - For each edge (i, j), the entries (i,j) and (j,i) are set to its weight,
+            or to 1.0 if the edge has no weight set.
+        - Node weights are stored in the diagonal since self-loops are not supported.
+            Missing node weights are set to 0.0 in the diagonal.
         """
-        nodes = sorted(self.nodes)
-        n_nodes = len(nodes)
-        index = {node: i for i, node in enumerate(nodes)}
-
-        matrix = np.zeros((n_nodes, n_nodes))
+        n_nodes = len(self.nodes)
+        index = {node: i for i, node in enumerate(self.nodes)}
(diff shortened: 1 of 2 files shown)


qoolqit/graphs/data_graph.py
```diff
@@ -294,6 +295,34 @@ def from_matrix(cls, data: np.ndarray) -> DataGraph:
         graph.edge_weights = edge_weights
         return graph
 
+    def to_matrix(self) -> npt.NDArray[np.float64]:
+        """Return the connectivity matrix of this graph.
```
alessandro-santini: I would call this matrix "adjacency matrix", I don't know if "connectivity matrix" is usually used

alessandro-santini (dismissed):
Seems fine to me! thanks!

qoolqit/graphs/data_graph.py
sgrava: Good point!
Changed to `"""Return the adjacency matrix of this graph.`
It fits better with networkx language.

alessandro-santini (approved):
fine for me thanks!

PR #438, opened 2026-08-07
fix edge weights does not reflect edges added after construction

### LINKED ISSUE #431: edge_weights does not reflect edges added after construction
`edge_weights` does not reflect edges added after construction

## Description
`DataGraph._edge_weights` is populated by `_reset_dicts()`, which is called only at construction time. Since `add_edges_from` is inherited from NetworkX and does not update `_edge_weights`, any edges added after construction are missing from edge_weights with no error or warning.

## Root cause
`BaseGraph._reset_dicts` seeds `_edge_weights` from `sorted_edges` once, at construction. Subsequent `add_edges_from `calls go directly to the NetworkX layer and bypass `_reset_dicts`.
Minimum reproducible example
```python
from qoolqit.graphs.data_graph import DataGraph

g = DataGraph.from_nodes([0, 1, 2])
g.add_edges_from([(0, 1), (1, 2)])

print(g.edge_weights)  # {} — expected {(0, 1): None, (1, 2): None}
```
Expected behaviour
`{(0, 1): None, (1, 2): None}`

sgrava:
Fix bug #431, where graph's weights were going out of sync when added after the creation of a graph.
Main fix is computing node coordinates and node/edge weights on the fly.

## Changes
1. removing `_rested_dict()` method and all its calls
2. removing internal private dicts `_coords`, `_node_weights`, _edge_weights``
3. `DataGraph` methods `coords`, `node_weights` and `edge_weights` now extract the attributes directly from the graph rather then relying on the internal dicts of point 2.

## Additional Changes
- removing redundant default constructors in `BaseGraph` and `DataGraph`.


abussy-pasqal (commented):
If I understand, data was duplicated between the inherited `nx.Graph` and internal dicts. And both were not syncronized correctly. So this PR removes the latter ?

sgrava:
> If I understand, data was duplicated between the inherited `nx.Graph` and internal dicts. And both were not syncronized correctly. So this PR removes the latter ?

Correct! It removes the internal duplicated (but not updated) dicts.

abussy-pasqal (approved):

PR #440, opened 2026-08-11
Fix int-truncation bug in Waveform array sampling

### LINKED ISSUE #439: Waveform.__call__ silently truncates float samples to int for np.ndarray input
## Symptom

```python
import numpy as np
from qoolqit import RampWaveform, ConstantWaveform

w1 = ConstantWaveform(1, 0)
w2 = RampWaveform(1, 0, -2)
w_tot = w1 >> w2

n = 20
T = 2
w_tot(np.linspace(0, T, n))          # [0,0,...,0,-1,-1,-1,-1,-2]   (WRONG)
w_tot([i*T/(n-1) for i in range(n)]) # [0,...,-0.105,...,-2.0]      (correct)
```

Calling a waveform with a numpy array of times gives integer-truncated garbage, while calling with an equivalent Python list of the same time values gives correct floats. In the repro, 9/20 samples are wrong, with values truncated toward zero (e.g. `-0.947` becomes `0`, `-1.368` becomes `-1`).

## Root cause (two compounding bugs)

**1. Type contract violation — `ConstantWaveform.function` doesn't cast to float**

`qoolqit/waveforms/waveforms.py`, `ConstantWaveform.function`:
```python
def function(self, t: float) -> float:
    return self.value
```
`self.value` is stored verbatim from the constructor. `ConstantWaveform(1, 0)` is built with the Python `int` `0`, so `function()` — and therefore `Waveform._single_call()` — returns an `int` even though the type hint promises `float`. Any subclass that forwards a stored/derived value without explicit float coercion is susceptible to the same leak (`RampWaveform`/`BlackmanWaveform` currently stay float only incidentally, via a true division in their arithmetic, not because it's enforced; `InterpolatedWaveform` is the only class that explicitly casts with `float(...)`).

**2. `np.vectorize` dtype inference amplifies the leak**

`qoolqit/waveforms/base_waveforms.py`, `Waveform.__call__`:
```python
def __call__(self, t):
    if isinstance(t, np.ndarray):
        return np.vectorize(self._single_call)(t)
    if isinstance(t, list):
        return [self._single_call(ti) for ti in t]
    return self._single_call(t)
```
`np.vectorize` without `otypes` calls the wrapped function once on the *first* element to infer the output dtype, then locks the whole result array to that dtype. Since the first sample (`t=0`) hits `ConstantWaveform.function` and returns `int(0)`, the entire output array — including later samples from the `RampWaveform` segment, which are genuinely fractional — gets cast to `int` and truncated. The list branch has no such inference step, so it's unaffected.

## Why this matters beyond the repro

Bug 2 means *any* waveform whose first sampled value happens to be an exact integer (0, 1, -2, ...) — regardless of bug 1 — will silently corrupt every other sample in an array call. Bug 1 means user-supplied int parameters (very natural, e.g. `ConstantWaveform(1, 0)`) can inject that int. Fixing only one bug leaves the other latent.

## Suggested fixes (minimal)

1. In `base_waveforms.py`, pin the vectorize output dtype: `np.vectorize(self._single_call, otypes=[float])(t)`.
2. Coerce to float once in `Waveform._single_call` so every subclass gets the guarantee for free: `return 0.0 if (...) else float(self.function(t))`.

sgrava:
## Summary
- Fixes #439: `np.vectorize` in `Waveform.__call__` inferred its output dtype from the first sample, silently truncating later float samples to int when that first value happened to be an int.
- Pins the vectorized output dtype to `float` and casts values to `float` at the points where they could leak as `int` (`_single_call`, `duration`, and the constructors of `ConstantWaveform`, `RampWaveform`, `BlackmanWaveform`).

## Test plan
- Added regression tests in `test_base_waveforms.py` and `test_derived_waveforms.py` covering the int-leak and the array/list sampling mismatch from the issue repro.
- Full test suite passes.

### DIFF
--- a/qoolqit/waveforms/base_waveforms.py
+++ b/qoolqit/waveforms/base_waveforms.py
@@ -53,7 +53,7 @@ class Waveform(ABC):
                 f"Extra arguments in {type(self).__name__} need to be passed as keyword arguments"
             )
 
-        self._duration = duration
+        self._duration = float(duration)
         self._params_dict = kwargs
 
         for key, value in kwargs.items():
@@ -99,7 +99,7 @@ class Waveform(ABC):
         return self._params_dict
 
     def _single_call(self, t: float) -> float:
-        return 0.0 if (t < 0.0 or t > self.duration) else self.function(t)
+        return 0.0 if (t < 0.0 or t > self.duration) else float(self.function(t))
 
     @overload
     def __call__(self, t: float) -> float: ...
@@ -109,7 +109,7 @@ class Waveform(ABC):
 
     def __call__(self, t: float | list[float] | np.ndarray) -> float | list[float] | np.ndarray:
         if isinstance(t, np.ndarray):
-            return np.vectorize(self._single_call)(t)
+            return np.vectorize(self._single_call, otypes=[float])(t)
         if isinstance(t, list):
             return [self._single_call(ti) for ti in t]
         return self._single_call(t)
--- a/qoolqit/waveforms/waveforms.py
+++ b/qoolqit/waveforms/waveforms.py
@@ -50,7 +50,9 @@ class RampWaveform(Waveform):
         initial_value: float,
         final_value: float,
     ) -> None:
-        super().__init__(duration, initial_value=initial_value, final_value=final_value)
+        super().__init__(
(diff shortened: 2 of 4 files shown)


v-vitale (approved):
Thanks for fixing this!

PR #445, opened 2026-08-12
fix(graphs): keep edge_weights/node_weights/coords in sync after add_edges_from

### LINKED ISSUE #431: edge_weights does not reflect edges added after construction

agu2347:
## What

`BaseGraph._reset_dicts()` seeds `_edge_weights` (and `_node_weights`) from the graph's current nodes/edges, but is only ever called at construction time (or explicitly, by the alternative constructors). `add_edges_from` is inherited directly from `nx.Graph` and never routes through it, so edges (and any nodes they implicitly create) added after construction are silently missing from `edge_weights`, with no error or warning:

```python
from qoolqit.graphs import DataGraph

g = DataGraph.from_nodes([0, 1, 2])
g.add_edges_from([(0, 1), (1, 2)])

print(g.edge_weights)  # {} -- expected {(0, 1): None, (1, 2): None}
```

This repo's own test suite was already routing around the bug -- `test_to_matrix_unweighted` called `graph._reset_dicts()` right after `add_edges_from` with a `# FIXME: ... see issue #431` comment above it.

## Fix

Override `add_edges_from` on `BaseGraph` (inherited by `DataGraph`, which doesn't define its own): call through to `nx.Graph.add_edges_from`, then seed any node/edge that's genuinely new with a weight/coordinate of `None`, matching `_reset_dicts()`'s own convention. Existing weights are left untouched, so this can't clobber `edge_weights`/`node_weights` set explicitly after construction.

New nodes are also handled (not just edges), since `add_edges_from` can implicitly introduce nodes that don't exist yet -- leaving those out of `_node_weights`/`_coords` would surface later as a `KeyError` in unrelated code paths (e.g. `to_matrix()`, `distances()`).

## Testing

- Removed the `_reset_dicts()` workaround (and its `# FIXME`) from `test_to_matrix_unweighted`.
- Added `test_add_edges_from_updates_weights_and_coords`, covering: the exact case from the issue, that existing weights survive a later `add_edges_from` call, and that edges introducing brand-new nodes extend `node_weights`/`coords` for those nodes too.
- Ran the full `test_base_graph.py` suite (47 tests): all pass except one pre-existing, unrelated failure (`test_from_nx_wrong_pos_attr`, a `&` vs `and` short-circuit bug in `from_nx`'s validation, confirmed present on `main` before this change too).
- Reverting just the `base_graph.py` change makes the new regression test fail with the exact symptom from the issue (`edge_weights == {}` instead of `{(0, 1): None, (1, 2): None}`).
- `ruff check` / `ruff format --check` clean on both changed files.

Fixes #431

### DIFF
--- a/qoolqit/graphs/base_graph.py
+++ b/qoolqit/graphs/base_graph.py
@@ -64,6 +64,40 @@ class BaseGraph(nx.Graph):
         self._node_weights = {n: None for n in self.nodes}
         self._edge_weights = {e: None for e in self.sorted_edges}
 
+    def add_edges_from(self, ebunch_to_add: Iterable, **attr: Any) -> None:
+        """Add all the edges in ebunch_to_add, keeping weight/coord dicts in sync.
+
+        `nx.Graph.add_edges_from` is used directly by the NetworkX layer and,
+        unlike the alternative constructors here, does not go through
+        `_reset_dicts()`. Left unhandled, edges (and any nodes they
+        implicitly create) added after construction are silently missing
+        from `_edge_weights`, `_node_weights` and `_coords` -- see #431.
+
+        Any newly-added node or edge is seeded with a weight/coordinate of
+        None, matching `_reset_dicts()`'s convention. Existing entries are
+        left untouched.
+
+        Arguments:
+            ebunch_to_add: container of edges, as accepted by NetworkX.
+            attr: edge data (or labels or objects) assigned via keyword
+                arguments, as accepted by NetworkX.
+        """
+        super().add_edges_from(ebunch_to_add, **attr)
+        if not hasattr(self, "_edge_weights"):
+            # Called from __init__ (via the initial `self.add_edges_from(edges)`),
+            # before `_reset_dicts()` has run for the first time. The constructor
+            # performs the initial sync itself right after, so there is nothing
+            # to do here yet.
+            return
+        for n in self.nodes:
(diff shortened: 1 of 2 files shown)


sgrava:
Hi @agu2347 ,
Thanks for proposing the bug fix! We really appreciate external inputs.
Unfortunately we were already fixing this in a separate PR #438 .

If you wish to contribute to the library, please consider other [open issues](https://github.com/pasqal-io/qoolqit/issues).
Thanks.

Issue #448, opened 2026-08-13
Drop the list input branch of edge_weights setter


sgrava:
## Description
`BaseGraph.edge_weights`'s setter accepts a `dict` or a `list`. The `list` case maps positionally against `self.sorted_edges`:

```python
weights_dict = {i: w for i, w in zip(self.sorted_edges, weights)}
```

Two independent problems:
1. `sorted_edges` returns a `set`; its iteration order isn't part of any contract, so a caller has no reliable way to know which list position maps to which edge.
2. Even with a deterministic order, it wouldn't be intuitive: the canonical order comes from sorting by node *value*, not from the order edges were added via `add_edges_from(...)`. A caller would have to inspect `sorted_edges` first to discover the order before building a correctly-positioned list — at which point they could just build a dict directly.

Contrast with `node_weights`/`coords`, whose `list` branches zip against `self.nodes` (an ordered view matching however the caller added nodes) — deterministic and intuitive.

## Evidence it's not needed
- No usage of `edge_weights = [...]` anywhere in the codebase, docs, or notebooks.
- `docs/fundamentals/graphs.ipynb` builds edge weights as a dict keyed by `graph.sorted_edges`, avoiding this path already.
- `nx.set_edge_attributes` doesn't support per-edge list assignment either — a bare list is treated as one scalar applied to every edge — so this isn't a shortcut for anything networkx provides.

## Proposal
Make `edge_weights`'s setter dict-only; raise `TypeError` for non-dict input.

## Related
#447 — same property area (edge ordering).

sgrava:
Closed in #450

Issue #447, opened 2026-08-13
edge_weights setter/getter rejects valid input due to forced edge-tuple ordering


sgrava:
## Description
`BaseGraph.edge_weights` (and the underlying `sorted_edges` property) require edge tuples to match a specific canonical orientation `(u, v)` with `u < v`, determined by sorting node values. In practice, callers don't know — and shouldn't need to know — which orientation networkx will report for a given edge, so a semantically valid input can be rejected outright.

## Reproduction
```python
from qoolqit.graphs.base_graph import BaseGraph

g = BaseGraph()
g.add_edges_from([(0, 1), (1, 2), (2, 0)])

g.edge_weights = {(0, 1): 0.3, (1, 2): 0.4, (2, 0): 0.5}
# ValueError: Set of edges in the given dictionary does not match the graph ordered edges.
```
The edge between nodes `0` and `2` is internally stored/reported as `(0, 2)`, so the otherwise-valid key `(2, 0)` fails the equality check in the setter (`base_graph.py`, `edge_weights.setter`).

## Impact
- False rejections: a caller who builds keys using the "wrong" (but equally valid, for an undirected graph) orientation gets a `ValueError` even though the edge exists.
- Crash on non-orderable node labels: `sorted_edges` calls `x[0] > x[1]` to pick a canonical orientation, which raises `TypeError` if node labels don't support `<` (e.g. custom objects). Since `BaseGraph` doesn't otherwise require nodes to be orderable, this is an unnecessary constraint.

## Notes
Root cause is that `sorted_edges`/`all_node_pairs` assume a total order on node labels to canonicalize `(u, v)` vs `(v, u)`. Fix needs to either canonicalize without relying on `<` (e.g. node insertion index, discussed as one option) or drop canonicalization and compare edges order-independently (e.g. via `frozenset`), while keeping in mind that the latter doesn't guarantee a predictable output orientation. Affects `edge_weights`, `min_distance`/`max_distance`, `ud_radius_range`.

PR #449, opened 2026-08-13
add base graph update tests

### LINKED ISSUE #446: improve base_graph coverage
Improve base_graph coverage

## Description
Following bug fix #431 #438 , a regression test was added.
Restructure base graph tests to check for all special attibutes `coords`, `node_weights` and `edge_weights`,  are kept up to date if the graph si modified.

## To add
- [ ] test empty BaseGraph
- [ ] test update `coords`, `node_weights` and `edge_weights`

sgrava:
Resolve #446 

Having a look at the coverage spawned two other small issues #447 #448.

## Changes
- test empty BaseGraph
- test update coords and weights

### DIFF
--- a/tests/test_graphs/test_base_graph.py
+++ b/tests/test_graphs/test_base_graph.py
@@ -101,10 +101,7 @@ def test_coords_update_extra_node() -> None:
 
 
 def test_node_coords_update() -> None:
-    graph = BaseGraph()
-
-    # add nodes without coordinates
-    graph.add_nodes_from([0, 1, 2])
+    graph = BaseGraph.from_nodes([0, 1, 2])
     assert graph.has_coords is False
 
     # set new coordinates
--- a/tests/test_graphs/test_base_graph.py
+++ b/tests/test_graphs/test_base_graph.py
@@ -69,6 +69,91 @@ def test_basegraph_init(n_nodes: int) -> None:
     assert len(graph.ud_edges(radius=10.0 * scale)) == max_n_edges
 
 
+def test_empty_graph() -> None:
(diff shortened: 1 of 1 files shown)


tests/test_graphs/test_base_graph.py
```diff
@@ -69,6 +69,91 @@ def test_basegraph_init(n_nodes: int) -> None:
...
+
+
+def test_node_coords_update() -> None:
+    graph = BaseGraph()
+
+    # nodes needs to be added before updating coordinates
+    with pytest.raises(
+        ValueError, match="Set of nodes in the given dictionary does not match the graph nodes."
+    ):
+        graph.coords = {0: (0.5, 0.5)}
```
abussy-pasqal: what is the state of the graph after the exception ? Is it ok to continue the test ?
I think that you shouldn't expect the graph to behave correctly after that. Hence you should move this case into a standalone test.

Same applies to the other tests

sgrava: good point, assignment/setting like this should simply fail beacuse the thing you want top set is simply not there

sgrava: maybe too much but better safe than sorry! Thanks for pointing this out

abussy-pasqal: my point was that each `with pytest.raises` should be in a separate test

sgrava: Ah I get the point now. Thanks, indeed each test looks better now.

abussy-pasqal (approved):

PR #450, opened 2026-08-13
Stop forcing edge order in edge_weights setter

### LINKED ISSUE #447: edge_weights setter/getter rejects valid input due to forced edge-tuple ordering

sgrava:
Fixes #447, #448.

Completely rewrite `edge_weights`. 
Impl suffered from two issues:
 1. Imposing edge order when not needed. Undirected graph does not have preferred edge order.
 2. Setting weights from list is not really useful because edges does not have an intuitive order to follow like nodes. User likely need edges anyway to construct weights.

## Changes
- Rewrite `edge_weights` setter.
    * check that edges are the same independently on the order
    * update edges direcly. Arguably more direct and understandable then `nx.set_edge_attribute(.., .. ,..)`
    * improve docstring clarity.

# Not fixed here
Leaves  `sorted_edges` flying around. 
- [ ] Open an issue to completely remove sorting as unnecessary.

### DIFF
--- a/qoolqit/graphs/base_graph.py
+++ b/qoolqit/graphs/base_graph.py
@@ -275,24 +275,13 @@ class BaseGraph(nx.Graph):
         return {(u, v): w for u, v, w in self.edges(data="weight")}
 
     @edge_weights.setter
-    def edge_weights(self, weights: list | dict) -> None:
-        """Set the dictionary of edge weights.
+    def edge_weights(self, weights: dict) -> None:
+        """Sets edge weights from a given dictionary of values.
 
         Arguments:
-            weights: list or dictionary of weights.
+            weights: a dictionary of edge weights.
         """
-        if isinstance(weights, list):
-            if len(weights) != self.number_of_edges():
-                raise ValueError("Size of the weights list does not match the number of nodes.")
-            weights_dict = {i: w for i, w in zip(self.sorted_edges, weights)}
-        elif isinstance(weights, dict):
-            edges = set(weights.keys())
-            if set(self.sorted_edges) != edges:
-                raise ValueError(
-                    "Set of edges in the given dictionary does not match the graph ordered edges."
-                )
-            weights_dict = weights
-        nx.set_edge_attributes(self, weights_dict, "weight")
+        nx.set_edge_attributes(self, weights, "weight")
 
     @property
     def coords(self) -> dict:
--- a/qoolqit/graphs/base_graph.py
+++ b/qoolqit/graphs/base_graph.py
@@ -276,14 +276,37 @@ class BaseGraph(nx.Graph):
 
     @edge_weights.setter
     def edge_weights(self, weights: dict) -> None:
-        """Sets edge weights from a given dictionary of values.
+        """Set the dictionary of edge weights.
+
+        Checks that the set of edges in the given dictionary matches the graph's edges.
+        Each edge may be keyed as (u, v) or (v, u) since the graph is undirected.
+
+        To partially update edge weights, use the attribute-like access pattern:
+        ```python
+        graph.edges[0,1]["weight"] = 0.5  # Update weight of edge (0,1)
+        ```
 
         Arguments:
             weights: a dictionary of edge weights.
 
         Raises:
-            KeyError: if an edge in the given weights dictionary is not found in the graph.
+            ValueError: if the set of edges in the given dictionary does not match
+                the graph's edges.
+
+        Example:
+            >>> graph = BaseGraph([(0, 1), (1, 2), (2, 0)])
+            >>> graph.edge_weights = {(0, 1): 0.3, (1, 2): 0.4, (2, 0): 0.5}
+            >>> graph.edge_weights
+            {(0, 1): 0.3, (0, 2): 0.5, (1, 2): 0.4}
         """
+        # frozenset canonicalizes each edge regardless of orientation, e.g.
+        # frozenset((u, v)) == frozenset((v, u)).
+        given = {frozenset(e) for e in weights}
+        expected = {frozenset(e) for e in self.edges}
+        if given != expected:
+            raise ValueError(
+                "Set of edges in the given dictionary does not match the graph's edges."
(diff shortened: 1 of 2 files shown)


abussy-pasqal (commented):
Can you add in the `BaseGraph` doc whether the graph is directed or not ?

qoolqit/graphs/base_graph.py
```diff
@@ -275,24 +275,40 @@ def edge_weights(self) -> dict:
...
+        # frozenset canonicalizes each edge regardless of orientation, e.g.
+        # frozenset((u, v)) == frozenset((v, u)).
+        given = {frozenset(e) for e in weights}
+        expected = {frozenset(e) for e in self.edges}
+        if given != expected:
+            raise ValueError(
+                "Set of edges in the given dictionary does not match the graph's edges."
+            )
+        for (u, v), w in weights.items():
+            self.edges[u, v]["weight"] = w
```
abussy-pasqal, line 314: if I understand correctly, `nx.set_edge_attributes` requires the proper ordering, but `[u, v]` doesn't ?

sgrava: Both do not require ordering on a nx.Graph which is undirected ( graph.edges[0,1] and graph.edges[1,0] are really  pointing to the same thing).
Additionaly, which is what we don't like here, `nx.set_edge_attributes` ignore missing keys, but her we want to set all edges.

```python
import networkx as nx

G = nx.Graph()
G.add_nodes_from([1, 2])  # The graph only has nodes 1 and 2

# The dictionary contains node 99, which DOES NOT exist in the graph
attrs = {
    1: {"color": "red"},
    2: {"color": "blue"},
    99: {"color": "green"}  # Missing key/node in the graph
}

# Native execution: this does NOT raise a KeyError
nx.set_node_attributes(G, attrs)

print(G.nodes[1])   # Output: {'color': 'red'}
print(99 in G)      # Output: False (node 99 was safely ignored)
```

sgrava:
> Can you add in the `BaseGraph` doc whether the graph is directed or not ?

Fair point, should be more clear now. Please, have a look.

qoolqit/graphs/base_graph.py
```diff
@@ -275,24 +278,40 @@ def edge_weights(self) -> dict:
...
+
+        Raises:
+            ValueError: if the set of edges in the given dictionary does not match
+                the graph's edges.
+
+        Example:
+            >>> graph = BaseGraph([(0, 1), (1, 2), (2, 0)])
+            >>> graph.edge_weights = {(0, 1): 0.3, (1, 2): 0.4, (2, 0): 0.5}
+            >>> graph.edge_weights
+            {(0, 1): 0.3, (0, 2): 0.5, (1, 2): 0.4}
```
abussy-pasqal, line 303:
```
the set of edges in the given dictionary does not match
                the graph's edges.
```
I would rather say that weights are not specified for all edges.

It would yield
```
        """Set the dictionary of edge weights.

        Each edge may be keyed as (u, v) or (v, u) since the graph is undirected.

        Weights must be specified for all edges, otherwise a ValueError is raised.
        To partially update edge weights, use the attribute-like access pattern:
        ```python
        graph.edges[0,1]["weight"] = 0.5  # Update weight of edge (0,1)
        ```

        Arguments:
            weights: list or dictionary of weights.
            weights: a dictionary of edge weights.

        Raises:
            ValueError: if the set of edges in the given dictionary does not match
                the graph's edges.

        Example:
            >>> graph = BaseGraph([(0, 1), (1, 2), (2, 0)])
            >>> graph.edge_weights = {(0, 1): 0.3, (1, 2): 0.4, (2, 0): 0.5}
            >>> graph.edge_weights
            {(0, 1): 0.3, (0, 2): 0.5, (1, 2): 0.4}
```

abussy-pasqal: With that, you can keep
```
Raises:
            ValueError: if the set of edges in the given dictionary does not match
                the graph's edges.
```
which is more precise, but at least the distinction all vs partial is made clear before

sgrava: Added what you suggested 
`Weights must be specified for all edges, otherwise a ValueError is raised.`

abussy-pasqal (approved):

Issue #455, opened 2026-08-18
set default compilation


sgrava:
Change default compilation strategy to just perform unit conversion.

The current default MAX_ENERGY profile should be set explicitly as
```python
program.compile_to(device=..., profile="max_energy")
```
or as
```python
from qoolqit.execution import CompilerProfile
program.compile_to(device=..., profile=CompilerProfile.MAX_ENERGY)
```

## To do
- [x] Allow string aliases in CompilerProfile
- [x] Set compilation profile explicitly in tests and notebooks
- [x] Set new default compilation DEFAULT. Former WORKING_POINT to deprecate.
- [ ] update documentation to reflect the new default

PR #456, opened 2026-08-18
Allow string aliases for CompilerProfile

sgrava:
## Summary
- `QuantumProgram.compile_to` now accepts `profile` as either a `CompilerProfile` enum member or its string alias (e.g. `"max_energy"`, `"working_point"`), converting via `CompilerProfile(profile)` before compilation.
- Updated the `compile_to` docstring for accuracy (correct string values, complete `Raises` section) and clarity.
- Added test coverage for string-alias equivalence and invalid-string rejection in `test_compile_to.py`.

Progresses #455 by addressing its first sub-task ("Allow string aliases in CompilerProfile").

### DIFF
--- a/qoolqit/program.py
+++ b/qoolqit/program.py
@@ -84,39 +84,48 @@ class QuantumProgram:
     def compile_to(
         self,
         device: Device,
-        profile: CompilerProfile = CompilerProfile.MAX_ENERGY,
+        profile: str | CompilerProfile = CompilerProfile.MAX_ENERGY,
         device_max_duration_ratio: float | None = None,
     ) -> None:
         """Compiles the quantum program for execution on a specific device.
 
-        The compilation process adapts the program to the device's constraints while
-        preserving the relative ratios of the original program parameters. Different
-        compilation profiles optimize for specific objectives:
+        The compilation process translates a program to make it runnable on a specific device:
 
-        - CompilerProfile.MAX_ENERGY (default): Scales the program to utilize the device's
+        - Dimensionalization: Rescale the drive amplitude and the register positions to
+            physical units compatible with the device.
+        - Translation: Translate the program to a lower-level representation (Pulser) that
+            can be executed on the device.
+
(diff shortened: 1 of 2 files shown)


qoolqit/program.py
```diff
@@ -84,39 +84,48 @@ def __repr__(self) -> str:
     def compile_to(
         self,
         device: Device,
-        profile: CompilerProfile = CompilerProfile.MAX_ENERGY,
+        profile: str | CompilerProfile = CompilerProfile.MAX_ENERGY,
```
abussy-pasqal:
```suggestion
        profile: Literal["max_energy", "working_point"] | CompilerProfile = CompilerProfile.MAX_ENERGY,
```

abussy-pasqal: - Note that you need to include
```python
from typing import Literal
```
- I like to start the `Literal` with the default value, but this is just a personal preference
- You will need to add a `type-ignore` or something similar in your invalid string test to satisfy `mypy`

qoolqit/program.py
```diff
@@ -84,39 +84,48 @@ def __repr__(self) -> str:
...
 
-        - CompilerProfile.MAX_ENERGY (default): Scales the program to utilize the device's
+        - Dimensionalization: Rescale the drive amplitude and the register positions to
+            physical units compatible with the device.
+        - Translation: Translate the program to a lower-level representation (Pulser) that
+            can be executed on the device.
+
+        There are two compilation profiles:
+
+        - "max_energy" (default): Scale the program to utilize the device's
```
abussy-pasqal: the default is not the string but the enum

qoolqit/program.py
sgrava: Good point, thanks!

qoolqit/program.py
sgrava: removed here and added correct default in `Args:`

abussy-pasqal (approved):

Issue #464, opened 2026-09-08
Enforce docstring presence/formatting across the library


sgrava:
## Motivation

We currently reformat existing docstrings (`pydocstringformatter` in pre-commit) but don't enforce that public modules, classes, and functions *have* a docstring, or that they follow a consistent style. Coverage is inconsistent across the codebase. We should lint for this the same way we lint for style/imports today.

As a related cleanup, `[tool.ruff.lint] extend-ignore = ["F841"]` (unused local variable) should be removed — it was masking a real category of bugs and only suppresses one instance today (`tests/test_program.py:143`), which is trivial to fix.

## Scope check

Running `ruff check --select D` against `qoolqit/` today reports **168 violations** (mostly `D102`/`D103`/`D107` undocumented methods/functions/`__init__`, plus some `D100`/`D101`/`D104` for modules/classes/packages). This is spread across nearly every module, so enabling the rule set outright will not pass CI without a large one-shot cleanup.

**Tests should be exempt** — docstrings on test functions add no value and would just be boilerplate noise (`test_foo` is documentation enough).

### Option A — `ruff` (`pydocstyle` rules), recommended

Ruff already reimplements `pydocstyle` natively (`D` rule codes) and is already our sole linter (pre-commit + CI), so this adds zero new tooling:

```toml
[tool.ruff.lint]
select = ["E", "F", "I", "Q", "D"]

[tool.ruff.lint.pydocstyle]
convention = "google"

[tool.ruff.lint.per-file-ignores]
"tests/**/*" = ["D"]
"*.ipynb" = ["E402", "I002", "D100", "D104"]
```

Pros: no new dependency, one linter to configure/run/keep updated, `--fix` auto-fixes a handful of formatting codes (blank lines, first-line summary, etc.). Cons: ruff's `D` implementation occasionally lags behind upstream `pydocstyle` for edge cases.

<details>
<summary>Option B - `flake8` + `flake8-docstrings` (wraps `pydocstyle`)</summary>

### Option B - `flake8` + `flake8-docstrings` (wraps `pydocstyle`)

Add `flake8` and `flake8-docstrings` as dev/pre-commit dependencies, configure via `.flake8`/`setup.cfg` (flake8 doesn't read `pyproject.toml` natively):

```ini
[flake8]
docstring-convention = google
extend-exclude = tests/*
```

[...]

PR #470, opened 2026-09-11
Expand ruff lint rule selection

sgrava:
## Summary
- Widen `[tool.ruff.lint] select` from `E, F, I, Q` to also include `UP`, `B`, `SIM`, with a documented `ignore` list for the rules left for separate discussion (including `B905`, `zip()` without `strict=`).
- Fix the resulting findings that were addressed: unused loop variables, a redundant `isinstance` check, `dict.keys()` iteration, and several `pyupgrade`/typing modernizations (`Optional[X]` -> `X | None`, `typing.Callable` -> `collections.abc.Callable`, etc).
- Drop `mccabe.max-complexity` and `flake8-quotes.docstring-quotes` — both configure rule sets (`C90`, and docstring quoting under `D`) that aren't in `select`, so they were dead config.

## Rationale
#464 treats `pyproject.toml`'s ruff config as the single source of truth for linting and is about to add the `D` (docstring) rule set on top of it. Before piling more rules onto that config, this PR turns on the other rule categories ruff already ships that we weren't using yet, fixes the findings that are genuine cleanups, and ignores the rest (including `B905`) for now — so the next docstring-focused PR lands on an intentionally-configured baseline instead of an implicit one.

## To do next
To avoid a big diff, this MR left some ignored rules to fix
```toml
ignore = [
    "B905", # `zip()` without an explicit `strict=` parameter
    "B008", # `np.random.default_rng` in argument defaults
    "B018", # Found useless expression
    "B024", # abstract base class with no abstract methods or properties
    "B904", # raise exceptions with `raise ... from err` or `raise ... from None`
    "SIM102", # Use a single `if` statement instead of nested `if` statements
    "SIM211", # Use `not ...` instead of `False if ... else True`
]
```
To be fixed in a separate issue #474

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,10 +107,25 @@ exclude_also = [
 line-length = 100
 
 [tool.ruff.lint]
-select = ["E", "F", "I", "Q"]
+select = [
+    "E", # pycodestyle
+    "F", # Pyflakes
+    "Q", # flake8-quotes
+    "UP", # pyupgrade
+    "B", # flake8-bugbear
+    "SIM", # flake8-simplify
+    "I", # isort
+]
+ignore = [
+    "B905", # `zip()` without an explicit `strict=` parameter
+    "B008", # `np.random.default_rng` in argument defaults
+    "B018", # Found useless expression
+    "B024", # abstract base class with no abstract methods or properties
+    "B904", # raise exceptions with `raise ... from err` or `raise ... from None`
+    "SIM102", # Use a single `if` statement instead of nested `if` statements
+    "SIM211", # Use `not ...` instead of `False if ... else True`
+]
 isort.required-imports = ["from __future__ import annotations"]
-mccabe.max-complexity = 15
-flake8-quotes.docstring-quotes = "double"
 
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401", "F403"]
--- a/docs/fundamentals/embedding.ipynb
+++ b/docs/fundamentals/embedding.ipynb
@@ -346,18 +346,7 @@
    "id": "27",
    "metadata": {},
    "outputs": [],
-   "source": [
-    "n=6\n",
-    "graph_2 = DataGraph.circle(n)\n",
-    "for i,e in enumerate(graph_2.edge_weights):\n",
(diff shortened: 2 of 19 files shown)


pyproject.toml
```diff
@@ -107,10 +107,25 @@ exclude_also = [
...
+    "F", # Pyflakes
+    "Q", # flake8-quotes
+    "UP", # pyupgrade
+    "B", # flake8-bugbear
+    "SIM", # flake8-simplify
+    "I", # isort
+]
+ignore = [
+    "B905", # `zip()` without an explicit `strict=` parameter
+    "B008", # `np.random.default_rng` in argument defaults
```
alessandro-santini, line 121: Ignoring b008 also hides possible unwanted features as the following which is raised as an error
```

from qoolqit.embedding import Blade

a = Blade()
a.config.max_min_dist_ratio = 99
b = Blade()
b.config.max_min_dist_ratio  # is 99
a.config is b.config         # True
```

pyproject.toml
```diff
@@ -107,10 +107,25 @@ exclude_also = [
...
+    "B018", # Found useless expression
```
alessandro-santini, line 122: We could keep B018 with the only fix of adding here

```
B018 Found useless expression. Either assign it to a variable or remove it.
  --> qoolqit/embedding/algorithms/blade/_force.py:84:9
   |
82 |         cleaned[self.vector_weights == 0] = 0.0
83 |         object.__setattr__(self, "weighted_vectors", cleaned)
84 |         self.maximum_temperatures 
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^
85 |
86 |     def get_nb_dims(self) -> int:
```

Changing only line 84 to 

```

self.maximum_temperatures  # noqa: B018 - precompute the cached property

```

and then we ignore B018 in all the tests

  ```
[tool.ruff.lint.per-file-ignores]
  "tests/**" = ["B018"]
```

alessandro-santini (approved):
Will be fix in a separate PR

pyproject.toml
sgrava: Added to #474

pyproject.toml
sgrava: Added tot #474

PR #473, opened 2026-09-16
Test docs portal preview

sgrava:
Test docs portal preview


.github/workflows/docs.yml
```diff
@@ -60,3 +63,76 @@ jobs:
...
+        needs: [deploy-docs-preview]
+        if: |
+          always() &&
+          github.event.pull_request.head.repo.full_name == github.repository &&
+          (github.event.action == 'closed' || needs['deploy-docs-preview'].result == 'success')
+        runs-on: ubuntu-latest
+        permissions:
+            contents: write
+            pull-requests: write
+            issues: write
```
abussy-pasqal, line 78: why permissions on `issues` ?

.github/workflows/docs.yml
```diff
@@ -60,3 +63,76 @@ jobs:
...
+        env:
+            PORTAL_PREVIEW_PATH: portal-preview/pr-${{ github.event.pull_request.number }}
+            DOC_PORTAL_IMAGE: europe-west9-docker.pkg.dev/pasqal-artifact/frontend/doc-portal:source-runtime
+        steps:
+            - name: Checkout the base commit, never the PR head
+              uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+              with:
+                  ref: ${{ github.event.pull_request.base.sha }}
+                  persist-credentials: false
```
abussy-pasqal, line 87: why isn't this protection applied to all actions/checkout with write permissions ?

.github/workflows/docs.yml
```diff
@@ -60,3 +63,76 @@ jobs:
...
+            - name: Pull the doc portal image
+              if: github.event.action != 'closed'
+              run: docker pull "$DOC_PORTAL_IMAGE"
+
+            - name: Build the portal preview
+              if: github.event.action != 'closed'
+              run: |
+                mkdir -p portal-dist
+                docker run --rm \
+                  --user 0:0 \
```
abussy-pasqal, line 106: why use root ?

.github/workflows/docs.yml
thomasbaronnet: You're right I removed issues: write, with the  sticky comments only need pull-requests: write.

.github/workflows/docs.yml
thomasbaronnet: You're right, fixed, both write jobs now checkout base.sha.

.github/workflows/docs.yml
thomasbaronnet: Needed so the container can write to the mounted portal-dist folder.

.github/workflows/docs.yml
```diff
@@ -60,3 +70,76 @@ jobs:
...
+              with:
+                  source-dir: ./portal-dist/
+                  preview-branch: gh-pages
+                  umbrella-dir: portal-preview
+                  qr-code: false
+                  comment: false
+
+            - name: Comment the preview link on the PR
+              if: github.event.action != 'closed'
+              uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2.9.4
```
abussy-pasqal, line 132: why not version 3 ?

abussy-pasqal (approved):

.github/workflows/docs.yml
thomasbaronnet: Copied the pin from pasqal-cloud / QEK, I can bump to v3 if you prefer.

sgrava: Let's update it already I'd say. 
Renovate bot will try to bump it anyway so let's just do it.

abussy-pasqal: It's up to you. I approved anyway

Issue #474, opened 2026-09-17
Fix ignored ruff lint rules left over from #470


sgrava:
## Context

#470 widened the ruff lint rule selection but left the following rules ignored to avoid a big diff:

```toml
[tool.ruff.lint]
ignore = [
    "B905", # `zip()` without an explicit `strict=` parameter
    "B008", # `np.random.default_rng` in argument defaults
    "B018", # Found useless expression
    "B024", # abstract base class with no abstract methods or properties
    "B904", # raise exceptions with `raise ... from err` or `raise ... from None`
    "SIM102", # Use a single `if` statement instead of nested `if` statements
    "SIM211", # Use `not ...` instead of `False if ... else True`
]
```

## Tasks
Fix the underlying findings for these rules and remove them from the `ignore` list in `pyproject.toml`.

 - [ ] "B905", # `zip()` without an explicit `strict=` parameter
 - [x] "B008", # `np.random.default_rng` in argument defaults
 - [x] "B018", # Found useless expression
 - [ ] "B024", # abstract base class with no abstract methods or properties
 - [ ] "B904", # raise exceptions with `raise ... from err` or `raise ... from None`
 - [ ] "SIM102", # Use a single `if` statement instead of nested `if` statements
 - [ ] "SIM211", # Use `not ...` instead of `False if ... else True`

## To explore

Follow-ups from review comments on #470 that go beyond a mechanical lint fix:

- [x] **B008 — shared mutable default hides a real bug, not just a style nit.** `Blade.__init__(self, config: BladeConfig = BladeConfig())` in [`matrix_embedder.py`](qoolqit/embedding/matrix_embedder.py) uses a mutable default argument, so every `Blade()` instance created without an explicit `config` shares the *same* `BladeConfig` object. Mutating `a.config` therefore leaks into `b.config` for any other default-constructed `Blade`:
  ```python
  from qoolqit.embedding import Blade

  a = Blade()
  a.config.max_min_dist_ratio = 99
  b = Blade()
  b.config.max_min_dist_ratio  # is 99
  a.config is b.config         # True
  ```
  Fix by constructing the default inside `__init__` (e.g. `config: BladeConfig | None = None` then `config = config or BladeConfig()`) rather than continuing to ignore B008 across the board. Check for the same pattern elsewhere before just re-enabling the rule.
  (from [review comment](https://github.com/pasqal-io/qoolqit/pull/470#discussion_r4035037131))

[...]

PR #478, opened 2026-09-18
Remove B008 from ignore and fix the errors

alessandro-santini:
Removed B008 from the ignored ruff errors. Fixed the codebase accordingly.

Part of #474


pyproject.toml
```diff
@@ -130,7 +129,7 @@ isort.required-imports = ["from __future__ import annotations"]
 "__init__.py" = ["F401", "F403"]
 "*.ipynb" = ["E402", "I002"]
 "setup.py" = ["I002"]
-"tests/**" = ["B018"]
+"tests/**" = ["B008","B018"]
```
sgrava:
```suggestion
"tests/**" = ["B018"]
```

Let's keep it if it doesn't make the linter to complain.
Ignoring B018 is more like a special exception just for tests

sgrava (changes requested):

sgrava (approved):
Looks good!

PR #312, opened 2026-05-12
[Docs] Add blade() function to the documentation

abussy-pasqal:
* Rename with leading _
* Reformat docstring
* Add to internals.md
* Mark function as private in the doc

> Thank you for submitting a change to this project!
>
> If you haven't done so yet, we suggest you read the (see docs/CONTRIBUTOR AGREEMENT.md). This is a legal document that details the legal rights and obligations of both Pasqal and you regarding your submission. Whenever you submit code, you accept this contributor licence agreement.

PR #479, opened 2026-09-21
[Doc] Add docstrings to BladeConfig

abussy-pasqal:
Document each BladeConfig field inline instead of relying on the private _blade docstring, and remove the internals reference page that exposed _blade in the public docs.


abussy-pasqal:
Clean version of #312

sgrava (commented):
The `pydocstringformatter` pre-commit hook forces you to do one line summary, blank line and then body.
So the linter probably messed up with that. Some fixes.

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
     """Configuration parameters to embed with BLaDE."""
 
     max_min_dist_ratio: float | None = None
+    """If present, set the maximum ratio between.
+
+    the maximum radial distance and the minimum pairwise distances.
+    """
```
sgrava: pydocstringformatter split this mid-sentence (it splits after the first physical line, not the first full sentence) and left a fragment ending in a lone period. Suggested rewrite gives it a real one-line summary so it survives another pydocstringformatter run unchanged:

```suggestion
    """If present, sets the maximum radial-to-pairwise distance ratio.

    Bounds the ratio between the maximum radial distance and the minimum
    pairwise distances.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+
     dimensions: tuple[int, ...] = default_dimensions
+    """List of numbers of dimensions to explore one.
+
+    after the other. A list with one value is equivalent to a list containing
+    twice the same value. For a 2D embedding, the last value should be 2.
+    Increasing the number of intermediate dimensions can help to escape
+    from local minima.
+    """
```
sgrava: Same issue here — "explore one." is a mid-sentence fragment. Rewritten so the first line is a complete sentence:

```suggestion
    """List of numbers of dimensions to explore one after the other.

    A list with one value is equivalent to a list containing twice the same
    value. For a 2D embedding, the last value should be 2. Increasing the
    number of intermediate dimensions can help to escape from local minima.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+
     starting_positions: np.ndarray | None = None
+    """If provided, initial positions to start from.
+
+    Otherwise,
+    random positions will be generated. The number of dimensions of the
+    starting positions must be lower than or equal to the first dimension
+    to explore. If it is lower, it is added dimensions filled with
+    random values.
+    """
```
sgrava, line 659: Here the split leaves a dangling "Otherwise," on its own line right after the blank line. Suggest folding it back into the body paragraph:

```suggestion
    """If provided, initial positions to start from.

    Otherwise, random positions will be generated. The number of dimensions
    of the starting positions must be lower than or equal to the first
    dimension to explore. If it is lower, it is added dimensions filled with
    random values.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+
     pca: bool = default_pca
+    """Whether to apply Principal Component Analysis to prioritize dimensions.
+
+    to keep when transitioning from a space to a space with fewer dimensions.
+    It is disabled by default because it can raise an error when there are
+    too many dimensions compared to the number of nodes.
+    """
```
sgrava: "...to prioritize dimensions." isn't a complete thought on its own. Suggested summary line:

```suggestion
    """Whether to apply Principal Component Analysis to prioritize dimensions to keep.

    Applies when transitioning from a space to a space with fewer dimensions.
    It is disabled by default because it can raise an error when there are
    too many dimensions compared to the number of nodes.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+
     steps_per_round: int = default_steps_per_round
+    """Number of elementary steps to perform for each dimension.
+
+    transition, where at each step move vectors are computed and applied
+    on the nodes.
+    """
```
sgrava: "...for each dimension." cuts off before "transition", which changes the meaning. Suggested fix:

```suggestion
    """Number of elementary steps to perform for each dimension transition.

    At each step, move vectors are computed and applied on the nodes.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+    and 1 (full regulation) that uniformizes the ability for the forces
+    to achieve their objectives at each step by changing priorities.
+    """
+
     compute_ratio_step_factors: Callable[[float], float] = default_compute_ratio_step_factors
+    """Function that is called at the boundaries of.
+
+    the rounds. It defines the target ratio the enforce during the
+    evolution. It acts as a multiplying factor on the target ratio.
+    """
```
sgrava: "...at the boundaries of." is a broken sentence, and there's also a pre-existing typo ("ratio the enforce" -> "ratio to enforce") that I fixed while rewriting this one:

```suggestion
    """Function that is called at the boundaries of the rounds.

    It defines the target ratio to enforce during the evolution. It acts
    as a multiplying factor on the target ratio.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+
     ratio_rerun: int = default_ratio_rerun
+    """When the distance ratio constraint is not met, it defines.
+
+    the maximum number of times the algorithm applies additional
+    computation steps putting the priority on the constraint.
+    """
```
sgrava: Same mid-sentence split ("...it defines."). Suggested rewrite:

```suggestion
    """Maximum number of retries when the distance ratio constraint is not met.

    Defines how many additional computation steps the algorithm performs,
    putting priority on satisfying the constraint.
    """
```

qoolqit/embedding/algorithms/blade/blade.py
```diff
@@ -635,20 +635,97 @@ class BladeConfig(EmbedderConfig):
...
+
     device: InitVar[Device | None] = None
+    """The QoolQit device to use to set the maximum ratio between the maximum.
+
+    radial distance and the minimum pairwise distance between atoms.
+    """
```
sgrava: "...between the maximum." is cut off before it says what it's comparing to. Suggested rewrite:

```suggestion
    """The QoolQit device used to compute the maximum distance ratio.

    Determines the maximum ratio between the maximum radial distance and
    the minimum pairwise distance between atoms.
    """
```

sgrava (changes requested):

sgrava (approved):
Nice

